PR #31, opened 2019-08-29
Fix hyperparameter validation for objective when eval_metric is auc.

ericangelokim:
*Description of changes:*

Fixed hyperparameter validation to use black list of objective functions when eval_metric includes 'auc'.

Note: Need to double check for multiclass objectives

*Test*
python3 -m tox

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
@@ -94,8 +94,8 @@ def initialize(metrics):
     @hpv.dependencies_validator(["objective"])
     def eval_metric_dep_validator(value, dependencies):
         if "auc" in value:
-            if any(metric_type in dependencies["objective"] for metric_type in [
-                    'count:', 'multi:', 'reg:', 'survival:']):
+            if not any(metric_type in dependencies["objective"] for metric_type in [
+                    'binary:', 'rank:']):
                 raise exc.UserError("Metric 'auc' can only be applied for classification and ranking problems.")
 
     hyperparameters = hpv.Hyperparameters(
--- a/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
@@ -94,10 +94,9 @@ def initialize(metrics):
     @hpv.dependencies_validator(["objective"])
     def eval_metric_dep_validator(value, dependencies):
         if "auc" in value:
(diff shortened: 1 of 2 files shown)


aws-patlin:
Looks like you aren't testing for 'multi:' currently. Is this because you're waiting on confirmation from Hyunsu that AUC won't work with multiclass?

ericangelokim:
Yes.

src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
```diff
@@ -94,10 +94,9 @@ def eval_metric_range_validator(SUPPORTED_METRIC, metric):
     @hpv.dependencies_validator(["objective"])
     def eval_metric_dep_validator(value, dependencies):
         if "auc" in value:
-            if (dependencies["objective"] not in
-                ["binary:logistic", "binary:logitraw", "multi:softmax", "multi:softprob",
-                 "reg:logistic", "rank:pairwise", "binary:hinge"]):
-                raise exc.UserError("Metric 'auc' can only be applied for classification and ranking problem.")
+            if not any(metric_type in dependencies["objective"] for metric_type in [
```
iyerr3: We might want to be little more specific by saying `dependencies["objective"].startswith(metric_type)`. It may not affect here since the objective will always start with `text:` but might be easier to understand and maintain.

ericangelokim: This is fair. I will change to starts with and squash merge.

src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
```diff
@@ -187,8 +186,8 @@ def eval_metric_dep_validator(value, dependencies):
         hpv.CategoricalHyperparameter(name="objective",
                                       range=["binary:logistic", "binary:logitraw", "binary:hinge",
                                              "count:poisson", "multi:softmax", "multi:softprob",
```
hcho3: I have some doubt about allowing multi-class classification objective when the AUC metric is set. To my knowledge, Area Under the [ROC Curve](https://en.wikipedia.org/wiki/Receiver_operating_characteristic) is only defined when the number of classes is 2. To plot ROC, we need to convert the multi-class classification problem into one-vs-all. See discussion at https://stats.stackexchange.com/questions/2151/how-to-plot-roc-curves-in-multiclass-classification

hcho3: Let me do some write-up about the definition of AUC

PR #32, opened 2019-09-09
Support checkpointing and reloading models

edwardjkim:
*Issue #, if available:*

*Description of changes:*

*Summary*
- Added `callback.SaveCheckpoint`: A callback for saving checkpoints.
- Added `train_utils.load_checkpoint`: For reloading models upon resuming training.
- Added unit tests in `test/unit/algorithm_mode/test_callback.py`.

*Tests*
- Unit tests: `pytest test/unit`
- Functional and integration tests: CR-12948549

*Context*

This PR is to support checkpointing for [managed spot training](https://docs.aws.amazon.com/sagemaker/latest/dg/model-managed-spot-training.html), which was recently launched. Since spot instances can be interrupted at anytime, we need to be able to save checkpoints during training, and we also need to be able to resume from the last checkpoint when the training job is restarted.

To use managed spot training, customers specify `EnableManagedSpotTraining` and `CheckpointConfig` in [CreateTrainingJob](https://docs.aws.amazon.com/sagemaker/latest/dg/API_CreateTrainingJob.html) or the corresponding parameters in the SDK:
```json
{
  "EnableManagedSpotTraining": true,
  "CheckpointConfig": { 
    "LocalPath": "/opt/ml/checkpoints",
    "S3Uri": "s3://bucket/key"
  }
}
```

When customers specify `CheckpoiintConfig`, the `LocalPath` becomes available in the config file at `/opt/ml/input/config/checkpointconfig.json`:
```json
{
  "LocalPath": "/opt/ml/checkpoints"
}
```

Any files saved at this location will automatically be uploaded to S3. When training job is resumed from interruption, all uploaded files will be downloaded to the same location before training restarts.

*Implementation*

We first check if the config file is present at `/opt/ml/input/config/checkpointconfig.json`. If there is no config file, we don't save checkpoints. If it is available, we instantiate `callback.SaveCheckpoint` and add it to the list of callback functions that are called at the end of each iteration.

`SaveCheckpoint` saves each checkpoint to different files at the end of each iteration by appending the iteration number to the file name, e.g., `xgboost-checkpont.1`, `xgboost-checkpont.2`, and so on. When files are written to the directory specified by `LocalPath`, SM will automatically upload all files to the S3 location specified by `S3Uri`.

Since saving one checkpoint per iteration could result in a large number of files to save in S3 and download when spot instances are resumed, we retain only the 5 most recent checkpoints in the directory. This is accomplished by a background thread that deletes all checkpoints older than 5 most recent checkpoints (the number of files to keep is somewhat arbitrary, choosing the optimal number of files to keep is left for future work). Note that when a file is being uploaded by SageMaker, SM will create a marker file (file name + `.sagemaker-uploading`) to indicate that the file is being uploaded. SM will also create another marker file (file name + `.sagemaker-uploaded`) when the upload is completed. Thus, the background will skip deleting a file and try again later if there is a marker file `<filename>.sagemaker-uploading` and only attempt to delete a file when the marker file `<filename>.sagemaker-uploaded` is present.

*CloudWatch logs*

Suppose the customer is training for 100 rounds, and the spot instance is interrupted after iteration 49. When the training job is resumed, we want to load `xgboost-checkpont.49` and restart training from iteration 50. We use the iteration numbers in the file names in our modified version of `print_evaluation()` function to provide logs that are consistent with what customers would expect.

First phase: Starting ↠ Downloading ↠ Training ↠ Interrupted
```
[17:41:02] 5845x9 matrix with 46760 entries loaded from /opt/ml/input/data/train
[17:41:02] 1252x9 matrix with 10016 entries loaded from /opt/ml/input/data/validation
INFO:root:Single node training.
INFO:root:Train matrix has 5845 rows
INFO:root:Validation matrix has 1252 rows
[0]#011train-rmse:8.11829#011validation-rmse:8.1177
...
[49]#011train-rmse:1.57065#011validation-rmse:1.91968
```

Second phase: Resumed ↠ Starting ↠ Downloading ↠ Training
```
[18:58:35] 5845x9 matrix with 46760 entries loaded from /opt/ml/input/data/train
[18:58:35] 1252x9 matrix with 10016 entries loaded from /opt/ml/input/data/validation
INFO:root:Single node training.
INFO:root:Checkpoint loaded from /opt/ml/checkpoints/xgboost-checkpoint.49
INFO:root:Resuming from iteration 50
[50]#011train-rmse:1.5635#011validation-rmse:1.91491
...
[99]#011train-rmse:1.30887#011validation-rmse:1.79031
```


src/sagemaker_xgboost_container/algorithm_mode/train_utils.py
```diff
@@ -11,8 +11,13 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 import logging
-
+import os
+import re
+from typing import Tuple, Optional
+import xgboost
```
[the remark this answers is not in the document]
edwardjkim: Done.

src/sagemaker_xgboost_container/training.py
```diff
@@ -37,12 +37,19 @@ def run_algorithm_mode():
...
+        'SM_CHECKPOINT_CONFIG_FILE'
 
     """
     # TODO: replace with CSDK constants in sagemaker_containers._env
     train_config = json.load(open(os.getenv(sm_env_constants.SM_INPUT_TRAINING_CONFIG_FILE), "r"))
     data_config = json.load(open(os.getenv(sm_env_constants.SM_INPUT_DATA_CONFIG_FILE), "r"))
 
+    checkpoint_config_file = os.getenv(sm_env_constants.SM_CHECKPOINT_CONFIG_FILE)
+    if os.path.exists(checkpoint_config_file):
+        checkpoint_config = json.load(open(checkpoint_config_file, "r"))
```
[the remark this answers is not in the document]
edwardjkim: I agree. I should have changed the previous lines instead of copying them. Changed.

test/unit/algorithm_mode/test_callback.py
```diff
@@ -0,0 +1,161 @@
...
+        env.rank = 0
+        env.iteration = 42
+        env.end_iteration = 100
+
+        callback = SaveCheckpoint(checkpoint_dir=self.test_dir)
+        callback(env)
+
+        file_path = os.path.join(self.test_dir, "xgboost-checkpoint.000000000042")
+        self.assertTrue(os.path.isfile(file_path))
+        self.assertTrue(len(os.listdir(self.test_dir)), 1)
```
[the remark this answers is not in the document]
edwardjkim: I have some basic unit tests for marker files in `test_SaveCheckpoint_uploading`. Integration tests don't really check for marker files.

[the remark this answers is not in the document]
aws-patlin: On that note, it would be good to include the information you've provided in this PR as comments in the code so it's clear for another developer in the future what you've done.

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+    training is complete, we put SENTINEL on the queue, and when we see the
+    SENTINEL, we clean up and exit the thread.
+    """
+
+    SENTINEL = None
+
+    def __init__(
+            self, checkpoint_dir: str, max_to_keep: int = 5, start_iteration: int = 0,
+            num_round: Optional[int] = None
+            ) -> None:
```
ericangelokim: Ending parens should be in the same line as last element, see pep: https://www.python.org/dev/peps/pep-0008/#indentation

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+FILE_SAFE_SUFFIX = ".sagemaker-uploaded"
+
+CHECKPOINT_FILENAME = "xgboost-checkpoint"
+CHECKPOINT_NUM_DIGITS = 12
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def _zero_pad(iteration: int) -> str:
```
ericangelokim: Method names should ideally start with an action, i.e. '_add_zero_pad' unless it is a property

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+
+    iter_num = format(iteration, "0{}".format(CHECKPOINT_NUM_DIGITS))
+
+    return iter_num
+
+
+# modified from https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/callback.py
```
ericangelokim: Making sure the main change is the addition of the method argument ```start_iteration```? Can you explain why this was required in comments?

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+        self.start()
+
+    def __call__(self, env: CallbackEnv) -> Callable[[CallbackEnv], None]:
+
+        if not os.path.exists(self.checkpoint_dir):
+            os.makedirs(self.checkpoint_dir)
+
+        return self.callback(env)
+
+    def fmt_path(self, i: int) -> str:
```
ericangelokim: What is i? Why are we zero_padding the file name?

src/sagemaker_xgboost_container/algorithm_mode/callback.py
ericangelokim: Can we rename this method name to print_checkpointed_evaluation?

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+    -------
+    >>> save_checkpoint = SaveCheckpoint("/opt/ml/checkpoints")
+    >>> xgboost.train(prams, dtrain, callbacks=[save_checkpoint])
+
+    At the end of each iteration, we save each checkpoint to different files
+    parametrized by the iteration number, e.g., checkpoint.01, checkpoint.02,
+    etc. However, this could result in a large number of files to save locally
+    (and on S3 if used in SageMaker). To save disk space and reduce the amount
+    of data required to be downloaded on resumption, we retain only the N
+    (default 5, spcified by 'max_to_keep') most recent checkpoints.
```
ericangelokim: Based on my understanding, if we have checkpoints queued up, the upload would only happen on the most recent checkpoint. So we retain n - 1 most recent uploaded checkpoints the queued checkpoints that have not been uploaded yet.

If my understanding is incorrect please let me know.

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+            return
+        i = env.iteration
+        if i % period == 0 or i + 1 == env.begin_iteration or i + 1 == env.end_iteration:
+            msg = '\t'.join([fmt_metric(x, show_stdv) for x in env.evaluation_result_list])
+            rabit.tracker_print('[%d]\t%s\n' % (i + start_iteration, msg))
+    return callback
+
+
+class SaveCheckpoint:
+    """Create a callback that saves checkpoints to disk.
```
ericangelokim: Minor: Doc reads like method documentation

src/sagemaker_xgboost_container/algorithm_mode/callback.py
ericangelokim: Also, if we are keeping this method, can you just rename it to format_path? You don't gain anything by removing the couple of letters, and you lose readability later.

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+    >>> xgboost.train(prams, dtrain, callbacks=[save_checkpoint])
+
+    At the end of each iteration, we save each checkpoint to different files
+    parametrized by the iteration number, e.g., checkpoint.01, checkpoint.02,
+    etc. However, this could result in a large number of files to save locally
+    (and on S3 if used in SageMaker). To save disk space and reduce the amount
+    of data required to be downloaded on resumption, we retain only the N
+    (default 5, spcified by 'max_to_keep') most recent checkpoints.
+
+    To delete stale checkpoints, we use a producer-consumer pattern: we start a
```
ericangelokim: This doc is nice, but we should also have comments near the code that actually executes it.

src/sagemaker_xgboost_container/algorithm_mode/callback.py
edwardjkim: Will change the variable and function names to be more descriptive.

On zero padding, I used the same format that tensorboard uses to save tfevent files (I just happened to be working on something else that uses tensorboard at the time). I think the rationale is by zero padding, it becomes string-sortable.

ericangelokim: If that is the case, can you describe the rational for how much zero padding you want to do and document? Sortability is nice, but if you have a limited number of zero padding it affects scaleability. In general I would rather opt for scaleability rather than sortability; also why would users have to sort the checkpoints?

edwardjkim: By scaleability, do you mean we might run out of zeros, like `.999999999999` -> `.000000000000`. Choosing 12 digits was somewhat arbitrary, and I don't expect anyone to train for trillion rounds.

Sortability is useful in `train_utils.load_checkpoint()` where I sort the checkpoints and try to load the latest checkpoint, but that can easily be changed if we remove zero-padding. I think you have a point; from the user's perspective, zero-padding is unnecessary.

It's a valid point, so I could get rid of zero-padding and modify the sorting method in `load_checkpoint()`. Do you think that's better?

ericangelokim: I agree with the trillion rounds, but no need to add limits that aren't necessary.

I think we should remove the zero-padding if there is no customer usecase for it.

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def start(self) -> None:
+
+        def _delete_once(skip_locked_files=True):
+
+            for i in iter(self.delete_queue.get, self.SENTINEL):
+
+                path = self.fmt_path(i)
+                if (skip_locked_files
+                        and os.path.isfile(path + FILE_LOCK_SUFFIX)
```
ericangelokim: Can you move these two ```.isfile()``` checks to a well named method for readability? I would also re-document the behavior here; whenever we have asynchronous processes I'd like to make sure its really clear what's happening.

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self.thread = threading.Thread(target=_delete_twice, daemon=True)
+        self.thread.start()
+
+    def stop(self) -> None:
+
+        self.delete_queue.put(self.SENTINEL)
+        self.thread.join()
+
+    def callback(self, env: CallbackEnv) -> None:
```
ericangelokim: Can you describe what is being done here? Between the file suffixes and trying to keep up with the file names its a bit confusing.

src/sagemaker_xgboost_container/algorithm_mode/callback.py
```diff
@@ -0,0 +1,205 @@
...
+        if env.rank != 0 or (not env.evaluation_result_list) or period is False or period == 0:
+            return
+        i = env.iteration
+        if i % period == 0 or i + 1 == env.begin_iteration or i + 1 == env.end_iteration:
+            msg = '\t'.join([fmt_metric(x, show_stdv) for x in env.evaluation_result_list])
+            rabit.tracker_print('[%d]\t%s\n' % (i + start_iteration, msg))
+    return callback
+
+
+class SaveCheckpoint:
```
ericangelokim: Overall this class seems much more complex than needed, or not complex enough. If we are deleting files based on 'flag' files why do we need to maintain a queue to remove files; wouldn't it be easier to just remove all files that we recognize as done? 

On the other side, I'm not as familiar with spot instances; what exactly is the behavior when a spot instance is taken away? Does the kernel recognize it as a interrupt? Is this process supposed to block on uploading the files, or do we just take it as best attempt and take whatever was managed to be saved in s3?

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,317 @@
...
+        that case, the file is put back on the queue and we try again later.
+        When training is complete, we put SENTINEL on the queue, and when we
+        see the SENTINEL, we clean up and exit the thread.
+        """
+        def _is_uploading(path):
+            uploading = os.path.isfile(path + FILE_LOCK_SUFFIX)
+            uploaded = os.path.isfile(path + FILE_SAFE_SUFFIX)
+            return uploading and not uploaded
+
+        def _delete_once(skip_locked_files=True):
```
ericangelokim: Why not just call this _delete_uploaded_files?

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,317 @@
...
+        return self.callback(env)
+
+    def format_path(self, iteration):
+        """Returns a file path to checkpoint given a iteration number"""
+        filename = "{}.{}".format(CHECKPOINT_FILENAME, iteration)
+        checkpoint_path = os.path.join(self.checkpoint_dir, filename)
+        return checkpoint_path
+
+    def start(self):
+        """Starts a background thread that deletes old checkpoints
```
ericangelokim: PEP 8 dictates that the docstring fgor methods read as a command, not a description:

```
"""Start background thread that ...
```

See more here: https://www.python.org/dev/peps/pep-0257/

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,317 @@
...
+        current_iteration = self.start_iteration + env.iteration
+        self._save_checkpoint(env.model, current_iteration)
+
+        # At the end of each iteration, we save each checkpoint to different files
+        # parametrized by the iteration number, e.g., checkpoint.1, checkpoint.2,
+        # etc. However, this could result in a large number of files to save locally
+        # (and on S3 if used in SageMaker). To save disk space and reduce the amount
+        # of data required to be downloaded on resumption, we retain only the N
+        # (default 5, spcified by 'max_to_keep') most recent checkpoints.
+        file_to_delete = self.format_path(current_iteration - self.max_to_keep)
```
ericangelokim: I'm personally not a fan of huge inline block comments; if you think a piece of code requires this level of commenting to make sense, it usually means the code is too confusing. I would consider refactoring to make it more clear, or even moving to its own helper method for clarity.

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,317 @@
...
+        file_to_delete_exists = os.path.isfile(file_to_delete)
```
ericangelokim: Instead of checking this here, I would just check during the actual delete. For example, just because the file exists now doesn't necessarily mean it will exist at deletion time, so this check isn't very protective.

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,317 @@
...
+        if file_to_delete_exists and (file_to_delete not in self.previous_checkpoints):
+            iteration_to_delete = current_iteration - self.max_to_keep
+            self.delete_queue.put(iteration_to_delete)
+
+        offset_iteration = env.end_iteration if self.num_round is None else self.num_round
+        training_has_ended = (current_iteration + 1 >= self.start_iteration + offset_iteration)
+        if training_has_ended:
+            self.stop()
```
ericangelokim: flake8 requires new line at end of file.

src/sagemaker_xgboost_container/training.py
```diff
@@ -37,11 +37,21 @@ def run_algorithm_mode():
...
     """
     # TODO: replace with CSDK constants in sagemaker_containers._env
-    train_config = json.load(open(os.getenv(sm_env_constants.SM_INPUT_TRAINING_CONFIG_FILE), "r"))
-    data_config = json.load(open(os.getenv(sm_env_constants.SM_INPUT_DATA_CONFIG_FILE), "r"))
+    with open(os.getenv(sm_env_constants.SM_INPUT_TRAINING_CONFIG_FILE), "r") as f:
+        train_config = json.load(f)
+    with open(os.getenv(sm_env_constants.SM_INPUT_DATA_CONFIG_FILE), "r") as f:
+        data_config = json.load(f)
+
+    checkpoint_config_file = os.getenv(sm_env_constants.SM_CHECKPOINT_CONFIG_FILE)
```
ericangelokim: Very nitpick, but I would just follow the pattern established above and skip the variable initialization for checkpoint_config_file.

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,315 @@
...
+        if env.rank != 0 or (not env.evaluation_result_list) or period is False or period == 0:
+            return
+        i = env.iteration
+        if i % period == 0 or i + 1 == env.begin_iteration or i + 1 == env.end_iteration:
+            msg = '\t'.join([format_metric(x, show_stdv) for x in env.evaluation_result_list])
+            rabit.tracker_print('[%d]\t%s\n' % (i + start_iteration, msg))
+    return callback
+
+
+def load_checkpoint(checkpoint_dir, max_try=5):
```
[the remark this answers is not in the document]
edwardjkim: Training won't start until after all the checkpoints have been downloaded, so we are safe there. The retries are for checking if we have corrupted files (e.g., partial uploads from a previous run).

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,317 @@
...
+        def _delete_twice():
+            _delete_once(skip_locked_files=True)
+            # Here, we've reached the end of training because we place sentinel in the
+            # queue at the end of training. We put another sentinel, go through everything
+            # in the queue once again, and try to remove it anyway where there is a lock on
+            # the file or not, because the training is done. On sagemaker, this should send
+            # a delete signal to the agent so that the upload can be canceled and removed
+            # from S3.
+            self.delete_queue.put(self.SENTINEL)
+            _delete_once(skip_locked_files=False)
```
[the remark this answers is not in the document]
edwardjkim: Yes, in the normal case, that function exits immediately and doesn't do anything.

I think my attempt to reuse the same function for two different things hurt readability here. I'm going to refactor and break `_delete_once()` into two functions even if they will share some code. I hope it improves readability.

src/sagemaker_xgboost_container/checkpointing.py
edwardjkim: I didn't know this. Good to know.

src/sagemaker_xgboost_container/checkpointing.py
edwardjkim: Thanks for this. I think it's a good suggestion.

src/sagemaker_xgboost_container/checkpointing.py
edwardjkim: Thanks for this. I think this is a good suggestion.

src/sagemaker_xgboost_container/checkpointing.py
edwardjkim: Is this true? When I include a new line, I get `W391 blank line at end of file`.

src/sagemaker_xgboost_container/checkpointing.py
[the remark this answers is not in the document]
edwardjkim: To add, the reason I'm going through everything twice is to avoid potential file corruption when we delete a file that is still being uploaded, because the EASE team was concerned about this aspect when I talked to them. Again, the second call won't do anything in normal cases. Added code comments to address this point, and hopefully the refactor makes code easier to read.

src/sagemaker_xgboost_container/checkpointing.py
ericangelokim: Have you run ```python3 -m tox```

test/unit/test_checkpointing.py
```diff
@@ -0,0 +1,230 @@
...
+import numpy as np
+import xgboost as xgb
+from sagemaker_xgboost_container import checkpointing
+from sagemaker_xgboost_container.checkpointing import SaveCheckpoint
+
+
+class TestSaveCheckpoint(unittest.TestCase):
+
+    def setUp(self):
+
```
ericangelokim: Super nitpick and personal choice, but you have a lot of unneeded new lines in this file :)

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,330 @@
...
+
+    callbacks = train_args.get("callbacks", [])
+    callbacks.append(print_checkpointed_evaluation(start_iteration=start_iteration))
+    callbacks.append(save_checkpoint(checkpoint_dir, start_iteration=start_iteration))
+
+    train_args["verbose_eval"] = False  # suppress xgboost's print_evaluation()
+    train_args["xgb_model"] = xgb_model
+    train_args["callbacks"] = callbacks
+    # xgboost's default value for num_boost_round is 10.
+    num_boost_round = train_args.get("num_boost_round", 10) - start_iteration
```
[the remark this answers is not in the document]
edwardjkim: Should we try to expose this constant upstream? We could do a PR that adds something like `DEFAULT_NUM_BOOST_ROUND = 10` in https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/training.py and replaces all `num_boost_round=10` with `num_boost_round=DEFAULT_NUM_BOOST_ROUND`.

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -0,0 +1,330 @@
...
+    # if last checkpoint is greater than num_boost_round, we shouldn't train.
+    train_args["num_boost_round"] = max(0, num_boost_round)
```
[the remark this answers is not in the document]
edwardjkim: We still need to call `xgb.train()` because `xgb_model` is a filename (str) while we have to return a booster object. But that line seems unnecessary. Removed it after testing `xgb.train()` with negative numbers and reading the source code upstream. Passing zero or negative number of rounds results in no training.

test/unit/test_checkpointing.py
edwardjkim: On retrying, the thought was there's a chance that some files are intact even if others are corrupt. We retain N=5 checkpoints (N retires is basically why we save N checkpoints), and I think it's much cheaper for users to save N checkpoints and try them all than to train from scratch if 1 fails.

Ran `python3 -m tox` and confirmed that all tests/flake8 pass.

edwardjkim:
Finished one final round of testing. Unit tests passed:
```
python3 -m tox
```
```
Required test coverage of 60% reached. Total coverage: 65.99%
py3-xgboost0.90: commands succeeded
flake8: commands succeeded
congratulations :)
```

Functional tests and integration tests passed. See CR-12948549.

I think it's ready to be merged.

PR #34, opened 2019-09-13
Added support for label_size=1 as an optional parameter for csv.

aws-patlin:
*Issue #, if available:*
https://issues.amazon.com/issues/AWSMLC-77

*Description of changes:*
Added additional handling in data_utils.py to check for 'text/csv' content type with the ``label_size`` optional parameter. If the ``label_size`` is present and is set to 1, the content type is accepted, otherwise a ``UserError`` is raised.

*Testing:*
Added two assert statements in the ``test_get_content_type`` unit test. Tests were run with tox and passed successfully.

### DIFF
--- a/src/sagemaker_xgboost_container/data_utils.py
+++ b/src/sagemaker_xgboost_container/data_utils.py
@@ -10,6 +10,7 @@
 # distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+import cgi
 import csv
 import logging
 import os
@@ -23,10 +24,12 @@ from sagemaker_xgboost_container.constants import xgb_content_types
 
 CSV = 'csv'
 LIBSVM = 'libsvm'
+VALID_CONTENT_TYPES = [CSV, LIBSVM, _content_types.CSV,
+                       xgb_content_types.LIBSVM, xgb_content_types.X_LIBSVM]
 
 
-INVALID_CONTENT_TYPE_ERROR = "{invalid_content_type} is not an accepted ContentType:" \
-                             " 'csv', 'libsvm', 'text/csv', 'text/libsvm', 'text/x-libsvm'."
+INVALID_CONTENT_TYPE_ERROR = "{invalid_content_type} is not an accepted ContentType: " + \
+                             ", ".join(['%s' % c for c in VALID_CONTENT_TYPES]) + "."
 INVALID_CONTENT_FORMAT_ERROR = "First line '{line_snippet}...' of file '{file_name}' is not " \
                                "'{content_type}' format. Please ensure the file is in '{content_type}' format."
 
@@ -43,13 +46,34 @@ def _get_invalid_csv_error_msg(line_snippet, file_name):
     return INVALID_CONTENT_FORMAT_ERROR.format(line_snippet=line_snippet, file_name=file_name, content_type='CSV')
 
 
+def _get_csv_content_type(content_type_cfg_val):
+    """
+    Return CSV if content_type_cfg_val is
+    * 'csv',
+    * 'text/csv',
+    * 'text/csv; ...' with valid parameters
+    """
+    if content_type_cfg_val in [CSV, _content_types.CSV]:
+        # Allow 'csv' and 'text/csv'
+        return CSV
+    else:
+        content_type, params = cgi.parse_header(content_type_cfg_val)
+        if content_type == _content_types.CSV:
+            if params.get('label_size') is not None and params['label_size'] != '1':
+                msg = "{} is not an accepted csv ContentType. "\
+                      "Optional parameter label_size must be equal to 1".format(content_type_cfg_val)
+                raise exc.UserError(msg)
+            return CSV
+    raise exc.UserError(_get_invalid_content_type_error_msg(content_type_cfg_val))
+
+
 def get_content_type(content_type_cfg_val):
     """Get content type from data config.
 
     Assumes that training and validation data have the same content type.
 
     ['libsvm', 'text/libsvm', 'text/x-libsvm'] will return 'libsvm'
-    ['csv', 'text/csv'] will return 'csv'
+    ['csv', 'text/csv', 'text/csv; label_size=1'] will return 'csv'
 
     :param content_type_cfg_val
     :return: Parsed content type
@@ -58,8 +82,8 @@ def get_content_type(content_type_cfg_val):
         return LIBSVM
     elif content_type_cfg_val.lower() in [LIBSVM, xgb_content_types.LIBSVM, xgb_content_types.X_LIBSVM]:
         return LIBSVM
-    elif content_type_cfg_val.lower() in [CSV, _content_types.CSV]:
-        return CSV
+    elif CSV in content_type_cfg_val.lower():
+        return _get_csv_content_type(content_type_cfg_val.lower())
     else:
         raise exc.UserError(_get_invalid_content_type_error_msg(content_type_cfg_val))
 
--- a/test/unit/test_data_utils.py
+++ b/test/unit/test_data_utils.py
@@ -31,9 +31,21 @@ class TestTrainUtils(unittest.TestCase):
 
         self.assertEqual('csv', data_utils.get_content_type('csv'))
         self.assertEqual('csv', data_utils.get_content_type('text/csv'))
+        self.assertEqual('csv', data_utils.get_content_type('text/csv; label_size=1'))
+        self.assertEqual('csv', data_utils.get_content_type('text/csv;label_size = 1'))
+        self.assertEqual('csv', data_utils.get_content_type('text/csv; charset=utf-8'))
+        self.assertEqual('csv', data_utils.get_content_type('text/csv; label_size=1; charset=utf-8'))
 
         with self.assertRaises(exc.UserError):
             data_utils.get_content_type('incorrect_format')
+        with self.assertRaises(exc.UserError):
+            data_utils.get_content_type('text/csv; label_size=5')
+        with self.assertRaises(exc.UserError):
+            data_utils.get_content_type('text/csv; label_size=1=1')
+        with self.assertRaises(exc.UserError):
+            data_utils.get_content_type('text/csv; label_size=1; label_size=2')
+        with self.assertRaises(exc.UserError):
+            data_utils.get_content_type('label_size=1; text/csv')
 
     def test_validate_csv_files(self):
         csv_file_paths = ['train.csv', 'train.csv.weights', 'csv_files']


aws-patlin:
Is it not a concern that user may include other optional parameters, such as 'text/csv; charset=utf-8'? If this happens, I think it makes sense for us to allow this, even if it is not officially supported.

ericangelokim:
I'm not saying we shouldn't do it, but we should wait to have the larger discussion with support for this across all containers.

aws-patlin:
Okay, in that case, is this bare-bones fix the right approach for now?

edwardjkim:
Do we allow `text/csv;label_size=1` (with no spaces after the semicolon)?. This [RCF](https://docs.aws.amazon.com/sagemaker/latest/dg/randomcutforest.html) documentation has `text/csv;label_size=1` and `text/csv;label_size=0 ` as examples. How about two (or more) spaces?

aws-patlin:
Yury notes that we should be as generally permissive of optional parameters and formatting. Reverted to a more general style of checking for optional parameters.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -60,6 +60,16 @@ def get_content_type(content_type_cfg_val):
         return LIBSVM
     elif content_type_cfg_val.lower() in [CSV, _content_types.CSV]:
         return CSV
+    elif _content_types.CSV in content_type_cfg_val.lower():
+        items = content_type_cfg_val.split(';')
+        label_size = [item for item in items if 'label_size' in item]
```
iyerr3: I would suggest renaming the variable to something similar to `all_label_sizes` to indicate that this is an iterable.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -60,6 +60,16 @@ def get_content_type(content_type_cfg_val):
...
+        if label_size:
+            label_size = label_size[0].split('=')[1]
```
iyerr3: Why is this only for the first label size? If we're making a generic solution, then might as well make it `any(label_size != 1 ... )`

test/unit/test_data_utils.py
```diff
@@ -31,9 +31,14 @@ def test_get_content_type(self):
...
         self.assertEqual('csv', data_utils.get_content_type('csv'))
         self.assertEqual('csv', data_utils.get_content_type('text/csv'))
+        self.assertEqual('csv', data_utils.get_content_type('text/csv; label_size=1'))
+        self.assertEqual('csv', data_utils.get_content_type('text/csv;label_size = 1'))
+        self.assertEqual('csv', data_utils.get_content_type('text/csv; charset=utf-8'))
 
         with self.assertRaises(exc.UserError):
             data_utils.get_content_type('incorrect_format')
+        with self.assertRaises(exc.UserError):
+            data_utils.get_content_type('text/csv; label_size=5')
```
iyerr3: how about `text/csv; label_size`, `text/csv; label_size=1=1` and `text/csv; label_size=1; label_size=2`?

aws-patlin: Good point, I wasn't thinking of these kinds of edge cases.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -60,6 +60,16 @@ def get_content_type(content_type_cfg_val):
         return LIBSVM
     elif content_type_cfg_val.lower() in [CSV, _content_types.CSV]:
         return CSV
+    elif _content_types.CSV in content_type_cfg_val.lower():
```
rizwangilani: From a readability perspective, I find this a little confusing as there is an explicit csv condition right above this (on line 61).
Would it make more sense to handle this in a single if condition (i.e. if its a csv) and work on label_size from there?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -60,6 +60,16 @@ def get_content_type(content_type_cfg_val):
...
+        items = content_type_cfg_val.split(';')
+        label_size = [item for item in items if 'label_size' in item]
+        if label_size:
+            label_size = label_size[0].split('=')[1]
+            if label_size.strip() != '1':
```
rizwangilani: How about making 1 as a parameter?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -60,6 +60,16 @@ def get_content_type(content_type_cfg_val):
...
+                msg = "{} is not an accepted csv ContentType. "\
+                      "Optional parameter label_size must be equal to 1".format(content_type_cfg_val)
```
rizwangilani: can reference the same parameter here

src/sagemaker_xgboost_container/data_utils.py
iyerr3: Good point - I think the lines 61-62 need to be removed.

src/sagemaker_xgboost_container/data_utils.py
rizwangilani: do we expect more than 1 label_size? is that a valid case?

aws-patlin: I feel like that should raise an exception, if multiple label_size values are given.

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: What would be the benefit in this case? By definition, label size for XGBoost must be 1, and this value is only referenced twice.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -43,13 +43,49 @@ def _get_invalid_csv_error_msg(line_snippet, file_name):
     return INVALID_CONTENT_FORMAT_ERROR.format(line_snippet=line_snippet, file_name=file_name, content_type='CSV')
 
 
+def _get_csv_content_type(content_type_cfg_val):
+    """
+    Returns CSV if content_type_cfg_val is
```
ericangelokim: Minor: PEP8, make this a command rather than desc ie ```Return CSV if content_type ...```

See examples here: https://www.python.org/dev/peps/pep-0257/

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -25,8 +25,8 @@
 LIBSVM = 'libsvm'
 
 
-INVALID_CONTENT_TYPE_ERROR = "{invalid_content_type} is not an accepted ContentType:" \
-                             " 'csv', 'libsvm', 'text/csv', 'text/libsvm', 'text/x-libsvm'."
+INVALID_CONTENT_TYPE_ERROR = "{invalid_content_type} is not an accepted ContentType: " \
```
ericangelokim: Minor. I would've expected ```text/csv;label_size=1``` without space. Did you check that this is the conventional pattern?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -43,13 +43,49 @@ def _get_invalid_csv_error_msg(line_snippet, file_name):
...
+    """
+    Returns CSV if content_type_cfg_val is
+    * 'csv',
+    * 'text/csv',
+    * 'text/csv; ...' with valid parameters
+    """
+    if content_type_cfg_val in [CSV, _content_types.CSV]:
+        # Allow 'csv' and 'text/csv'
+        return CSV
+    elif _content_types.CSV in content_type_cfg_val:
```
ericangelokim: This seems like a really ugly way to deal with a general problem. Have you looked into existing tools that can parse request header content types? such as: https://pypi.org/project/python-mimeparse/

It just feels like we are trying to reinvent the wheel here.

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: Yes, the public sagemaker documentation shows a space: https://docs.aws.amazon.com/sagemaker/latest/dg/cdf-training.html

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: Good point, I hadn't looked into existing tools.

aws-patlin: Looks like python has a library called 'cgi' with a [parse_header()](https://docs.python.org/2/library/cgi.html#cgi.parse_header) function which we can leverage. My only concern is that it handles duplicate parameters by keeping the last one (i.e. 'text/csv; label_size=1; label_size=2' produces a parameter dictionary with {'label_size': '2'}). Is this acceptable behavior?

aws-patlin: It also returns an empty dictionary if the parameter is only a name without a value (i.e. 'text/csv; label_size' -> `('text/csv', {})`), which would pass instead of erroring.

iyerr3: Does the standard for MIME/HTTP headers mention rules for options in content-type? 
If not then IMO it's better to be permissive than restrictive for optional parameters.

aws-patlin: The protocol doesn't have any guidelines regarding these edge cases: https://www.w3.org/Protocols/rfc1341/4_Content-Type.html
I'd say it should be fine to just go with the default behavior provided by cgi.parse_header()

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -25,8 +26,8 @@
 LIBSVM = 'libsvm'
 
 
-INVALID_CONTENT_TYPE_ERROR = "{invalid_content_type} is not an accepted ContentType:" \
-                             " 'csv', 'libsvm', 'text/csv', 'text/libsvm', 'text/x-libsvm'."
```
ericangelokim: Minor: These accepted ContentTypes should be put in a list so that we aren't copy/pasting raw string values. No need to submit a review for this change.

aws-patlin: Do you mean just for this error message, or for all references to these content types? I went ahead and created a list just for the error message, but it might be worthwhile to migrate everything into a dictionary (though this would be a separate issue to deal with).

ericangelokim: There should be a list that can be referenced throughout the codebase so that we aren't checking against a copy pasted list repeatedly. This isn't just for the error messages but throughout the repo.

aws-patlin: That would mean changing the way we're referencing these content types on a large scale. We should definitely do this, but I'm not sure this PR is the right place to make this change. What do you think?

ericangelokim: That's fair. Can we make the list for the error messages for now, since they refer to the same exact list? No need to resubmit for approval.

PR #36, opened 2019-10-07
Initial commit for pipe mode support.

aws-patlin:
*Description of changes:*
Added support for training with CSV pipe mode, parquet file and pipe mode, recordio-protobuf file and pipe mode.

Unit tests added for the new data reading methods, sagemaker-pipe.py from https://github.com/ishaaq/sagemaker-pipe added to emulate pipe mode during unit tests.

NOTE: Integration tests for checkpointing, inference, and HPO are failing when run on a local image. The error produced by the checkpointing tests are the same as before boto3 was updated to support `CheckpointConfig`, but pip freeze shows the latest version of boto3 installed in the container. HPO test is failing due to missing metric definition, which also doesn't make sense. Inference tests simply fail during endpoint creation. Not sure what is the cause of this problem, don't merge until resolved.

UPDATE: With an image built via CodeBuild, inference and HPO tests succeed, and checkpointing tests succeed after creating a new workspace, so these failures were not due to any change in this PR.


docker/0.90-1/base/Dockerfile.cpu
```diff
@@ -11,6 +11,17 @@ RUN cd /tmp && \
...
      rm get-pip.py
 
+# Install tensorio
+RUN echo 'installing miniconda'
+RUN curl -LO http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh
+RUN bash Miniconda3-latest-Linux-x86_64.sh -bfp /miniconda3
+RUN rm Miniconda3-latest-Linux-x86_64.sh
+ENV PATH=/miniconda3/bin:${PATH}
+RUN conda update -y conda
+RUN conda install protobuf tbb libiconv
```
cbalioglu: You don't need this line anymore. `mlio-py` installs them as its dependencies.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -288,7 +341,107 @@ def get_libsvm_dmatrix(files_path):
...
+            dmatrix = xgb.DMatrix(data[:, 1:], label=data[:, 0])
+            return dmatrix
+        else:
+            return None
+
+    except Exception as e:
+        raise exc.UserError("Failed to load parquet data with exception:\n{}".format(e))
+
+
+def get_recordio_protobuf_dmatrix(files_path):
```
cbalioglu: I assume that you are going to merge `get_recordio_protobuf_dmatrix` and `get_recordio_protobuf_dmatrix_pipe_mode` at some point.

docker/0.90-1/base/Dockerfile.cpu
aws-patlin: Great, thanks for catching that. I'll update the tox.ini as well.

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: I was going to leave them separate. Do you think it makes more sense to have the separate file mode and pipe mode functions all merged (i.e. only have one method per content type), or are you only pointing out recordio-protobuf in this case because the code is nearly identical?

rizwangilani:
I feel like there are a lot of changes for a single commit. Could we group the changes into smaller/more purposeful commits? perhaps move refactorization/indentation to a sep commit/etc? Let me know what you think.

aws-patlin:
> I feel like there are a lot of changes for a single commit. Could we group the changes into smaller/more purposeful commits?

Good point. Let me see if I can group this better into a couple commits. I'll separate it into build-related changes, implementation, and testing.

aws-patlin:
Commits have been split into Initial + CSV, Parquet, and RecordIO-Protobuf.

src/sagemaker_xgboost_container/algorithm_mode/channel_validation.py
```diff
@@ -19,20 +19,62 @@ def initialize():
     train_channel.add("csv", cv.Channel.FILE_MODE, cv.Channel.REPLICATED)
     train_channel.add("libsvm", cv.Channel.FILE_MODE, cv.Channel.SHARDED)
     train_channel.add("libsvm", cv.Channel.FILE_MODE, cv.Channel.REPLICATED)
+    train_channel.add("parquet", cv.Channel.FILE_MODE, cv.Channel.SHARDED)
```
ericangelokim: Minor: Let's loop the adds, instead of writing each one specifically,

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -288,7 +341,107 @@ def get_libsvm_dmatrix(files_path):
     return dmatrix
 
 
-def get_dmatrix(data_path, content_type, csv_weights=0):
+def get_parquet_dmatrix(files_path):
+    try:
+        table = pq.read_table(files_path)
+
+        data = table.to_pandas()
```
ericangelokim: Just a warning, but pandas is really slow. Can we avoid using pandas here?

test/unit/test_data_utils.py
```diff
@@ -79,6 +88,34 @@ def test_parse_csv_dmatrix(self):
 
                 self.assertEqual(no_weight_test_features, single_node_dmatrix.feature_names)
 
+    def test_parse_csv_dmatrix_pipe(self):
```
ericangelokim: The code for testing pipe mode has a lot of shared logic, can we abstract that out to a helper method here

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -299,22 +452,37 @@ def get_dmatrix(data_path, content_type, csv_weights=0):
...
             files_path = data_path
-        else:
+        elif not is_pipe:
             for root, dirs, files in os.walk(data_path):
                 if dirs == []:
                     files_path = root
                     break
         if content_type.lower() == CSV:
-            dmatrix = get_csv_dmatrix(files_path, csv_weights)
+            if is_pipe:
```
ericangelokim: Ideally this logic would look something like

reader = get_content_reader('csv') // this would return the correct reader for the format for example CSVReader for csv

dmatrix = get_dmatrix(reader, args, pipe)

I know that that requires changing code out of cope for this task, but can we at least add a todo?

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: It was mainly to follow the convention of the other functions, which are different enough to warrant two separate functions. In this case I'll combine it into a single function.

aws-patlin: What I could also do is have a single interface for both file mode and pipe mode functions, which actually sounds like a more appropriate design here.

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: Unfortunately this seems to be the only method of converting the pyarrow Table into a format that can be read by DMatrix. This is reflected by file mode training being slower than pipe mode. I'm not certain if there's a way to verify and enforce the conversion, but I will look into it.

test/unit/test_data_utils.py
aws-patlin: Will do.

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: There would need to be a few prerequisite changes in order for this to work:
- Use mlio for csv file mode 
- Have a data reader for libsvm (not yet supported)
- Have a data reader for parquet (the current ParquetRecordReader is actually an iterator over the records and does not follow the same interface as the other readers)

Should I leave a todo as a comment?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -238,13 +255,15 @@ def validate_data_file_path(data_path, content_type):
...
-
         if parsed_content_type.lower() == CSV:
             for data_file_path in data_files:
                 _validate_csv_format(data_file_path)
         elif parsed_content_type.lower() == LIBSVM:
             for data_file_path in data_files:
                 _validate_libsvm_format(data_file_path)
+        elif parsed_content_type.lower() == PARQUET or parsed_content_type.lower() == RECORDIO_PROTOBUF:
+            # No op
+            return
```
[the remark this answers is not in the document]
aws-patlin: Would a user of script mode ever use this method? I can add an else just in case, but in algo mode we only pass in a validated content type to this method.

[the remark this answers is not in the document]
aws-patlin: Actually, it looks like get_content_type() is called at the top of the function, so we're guaranteed that only valid content types will reach this part of the code. An else statement is not necessary.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -42,19 +43,22 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
     :param csv_weights: 1 if instance weights are in the second column of csv data files; otherwise, 0
```
ericangelokim: missing doc string for ```is_pipe```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -15,17 +15,34 @@
...
 import xgboost as xgb
 
 from sagemaker_algorithm_toolkit import exceptions as exc
 from sagemaker_containers import _content_types
 from sagemaker_xgboost_container.constants import xgb_content_types
 
+from mlio.integ.numpy import as_numpy
+from mlio.integ.arrow import as_arrow_file
+
+BATCH_SIZE = 4000
```
ericangelokim: Where does this come from?

test/utils/sagemaker_pipe.py
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python
+
+try:
+    # for python 3.x:
+    from urllib.parse import urlparse
+except ImportError:
+    # for python 2.x:
+    from urlparse import urlparse
```
ericangelokim: We are not running Python2 anymore, this shouldn't be needed

test/utils/sagemaker_pipe.py
```diff
@@ -0,0 +1,204 @@
...
+import gzip
+import logging
+import os
+import shutil
+import stat
+
+NUM_EPOCHS = 5
+
+
+def run(args):
```
ericangelokim: Why do none of these methods have doc string?

src/sagemaker_xgboost_container/algorithm_mode/train.py
aws-patlin: Resolved, also added doc strings for the new data utils.

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: I arbitrarily chose a value, since we need to specify a batch size for the mlio data readers. I have not experimented yet if memory consumption is affected by this value.

test/utils/sagemaker_pipe.py
aws-patlin: This was all copied from Ishaaq's [sagemaker pipe emulator](https://github.com/ishaaq/sagemaker-pipe/blob/master/sagemaker_pipe.py). I can try to add doc strings to these methods.

ericangelokim: Is there a reason this was copied? (Asking because I'm not as familiar with pipe mode).

If this does, please link original source in the file

aws-patlin: I copied this over in order to run it as a python script (which is how Ishaaq designed it to be used). I could also copy just the methods I need in order to create and delete the pipe, but I think this would be unnecessary work for little benefit.

ericangelokim: Ok. Please just refer to the original source in comments then!

PR #39, opened 2019-10-18
Mms integ [WIP]

ericangelokim:
*Issue #, if available:*

*Description of changes:*

Migrated both algo and script mode serving to MMS. The changes here can be clumped into the following:

* Dockerfile changes. Java is added for MMS, as well as copying configs and other artifacts to set up MMS.
* Serving.py has been stripped of most of the existing code to simply decide which MMS handler to initialize MMS with.
* Handlers, one for each script mode and algo mode. This has mostly been ported from the GUnicorn methods for (default_model_fn, default_input_fn, default_predict_fn, default_output_fn).

I am having trouble building with the master branch, so I still need deep dive on this issue. Has not been rebased with master yet.

Next steps:
1. Execution Parameters endpoint
2. Benchmark the speed of MMS v current implementation. The initialization of MMS is slower than before as we are starting a Java based server instead of GUnicorn, which is lighter. Need to make sure throughput has not been affected.


src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
```diff
@@ -0,0 +1,196 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aws-patlin: Is algorithm_mode/serve.py being kept? What is the difference between the two?

ericangelokim:
Removing in favor of rebased: https://github.com/aws/sagemaker-xgboost-container/pull/41

PR #41, opened 2019-10-30
Sagemaker XGB migrated to use MxNet Model Server

ericangelokim:
*Description of changes:*

Migrated to use MMS instead of the gunicorn server for inference.

Minor fixes to docs and requirements files.

This PR includes modified model-server.py and transformer.py files from the inference toolkit, as well as exception handling logic due to the latter not being available from upstream. Ideally these changes can be merged back into inference toolkit so as to not include duplicated effort in here.

# Tests Run
```tox``` tox test coverage at 64%
```pytest test/integration``` all tests passing
All container tests and integration tests in pipeline passing in local workspace.

### CODE REVIEW

**Comment by ericangelokim** [comment]
Removing https://github.com/aws/sagemaker-xgboost-container/pull/39 in favor of this PR.

File: setup.py [hunk 1]
```diff
@@ -37,5 +37,9 @@ def read(fname):
         'test': read("test-requirements.txt")
     },
 
+    entry_points={
+        'console_scripts': 'serve=sagemaker_pytorch_serving_container.serving:main'
```
edwardjkim: Why does it say pytorch?

src/sagemaker_xgboost_container/mms_patch/model_server.py
```diff
@@ -0,0 +1,178 @@
...
+def _adapt_to_mms_format(handler_service):
+    """Archive initial model using MMS handler
+
+    :param handler_service:
+    :return:
+    """
+    if not os.path.exists(DEFAULT_MMS_MODEL_DIRECTORY):
+        os.makedirs(DEFAULT_MMS_MODEL_DIRECTORY)
+
+    model_archiver_cmd = ['model-archiver',
```
vdantu: Why do you need to run model-archiver? I would recommend not using this and instead using ```default_service_handler``` (which is already used in the config.properties). The model-archiver command might add to the initialization latency, as it tries to zip or tar all the artifacts.

docker/0.90-1/final/Dockerfile.cpu
```diff
@@ -11,19 +56,13 @@ ENV SM_CHECKPOINT_CONFIG_FILE $SM_INPUT/config/checkpointconfig.json
...
-RUN pip install -r /requirements.txt && rm /requirements.txt
-
-# DMLC patches; TODO: remove after making contributions back to xgboost for tracker.py
-COPY src/sagemaker_xgboost_container/dmlc_patch/tracker.py \
-   /usr/local/lib/python3.5/dist-packages/xgboost/dmlc-core/tracker/dmlc_tracker/tracker.py
-
-COPY dist/sagemaker_xgboost_container-1.0-py2.py3-none-any.whl /sagemaker_xgboost_container-1.0-py2.py3-none-any.whl
-RUN pip install --no-cache /sagemaker_xgboost_container-1.0-py2.py3-none-any.whl && \
-    rm /sagemaker_xgboost_container-1.0-py2.py3-none-any.whl
+EXPOSE 8080 8081
```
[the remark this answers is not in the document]
ericangelokim: No; I will remove this port expose.

docker/0.90-1/resources/mms/ExecutionParameters.java
```diff
@@ -0,0 +1,79 @@
...
+        rsp.getOutputStream()
+                .write(
+                        new GsonBuilder()
+                                .setPrettyPrinting()
+                                .create()
+                                .toJson(r)
+                                .getBytes(StandardCharsets.UTF_8));
+    }
+
+    /** Response for Model server endpoint */
```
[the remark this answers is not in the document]
ericangelokim: This version parses environment variables. If that is ok, I can make this into an upstream change into MMS as well if given the correct env variable names to use.

docker/0.90-1/resources/mms/ExecutionParameters.java
```diff
@@ -0,0 +1,79 @@
+package software.amazon.ai.mms.plugins.endpoint;
```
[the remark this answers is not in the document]
ericangelokim: I am not actually sure. I can try removing it and re-running tests if required.

requirements.txt
```diff
@@ -1,12 +1,15 @@
 PyYAML<4.3
 matplotlib
+model-archiver
```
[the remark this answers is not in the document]
ericangelokim: Yes; in single model mode we use the archiver.

Actually, if I had more time I think I can get this working without running the archiver command, but I had trouble simply removing it and didn't want to spend too much time on this.

src/sagemaker_xgboost_container/algorithm_mode/mms_transformer.py
```diff
@@ -0,0 +1,174 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
ericangelokim: Good call out; I will add to all files.

src/sagemaker_xgboost_container/mms_patch/model_server.py
ericangelokim: Right. I actually tried running this without and I was having some trouble last week. I am ok revisiting this, but I kept it for now because eventually I wanted to merge these changes to upstream sagemaker-inference-toolkit which does use this; I'm not sure what I would break if I simply removed it there.

src/sagemaker_xgboost_container/algorithm_mode/mms_transformer.py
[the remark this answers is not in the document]
ericangelokim: Actually, I realized it's only added to python files. I think that it's ok that the copyright is not on the other (requirements, Dockerfiles, config, etc).

docker/0.90-1/resources/mms/ExecutionParameters.java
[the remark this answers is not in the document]
vdantu: You already have ```endpoints.jar``` included in your repository. That should technically contain this file within it. So, unless you are tracking this for your own purpose, I don't see a value of this file :) . 

On another note, you had some good findings and fixes. I would recommend you upstream that to the MMS repository.

docker/0.90-1/resources/mms/ExecutionParameters.java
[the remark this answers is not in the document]
vdantu: 👍 for upstreaming. Just make these generic and upstream them to MMS.

requirements.txt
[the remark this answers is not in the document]
vdantu: Ok

src/sagemaker_xgboost_container/mms_patch/model_server.py
vdantu: Sure. Makes sense.

docker/0.90-2/resources/mms/config.properties
```diff
@@ -0,0 +1,10 @@
+vmargs=-Xmx128m -XX:-UseLargePages -XX:+UseG1GC -XX:MaxMetaspaceSize=32M -XX:MaxDirectMemorySize=10m -XX:+ExitOnOutOfMemoryError
+model_store=$$SAGEMAKER_MMS_MODEL_STORE$$
+load_models=$$SAGEMAKER_MMS_LOAD_MODELS$$
+plugins_path=/tmp/plugins
+inference_address=http://0.0.0.0:$$SAGEMAKER_BIND_TO_PORT$$
+management_address=http://0.0.0.0:$$SAGEMAKER_BIND_TO_PORT$$
+default_workers_per_model=$$SAGEMAKER_NUM_MODEL_WORKERS$$
```
ChuyangDeng: With MME mode, the default_workers_per_model should be 1. Or are you forcing SAGEMAKER_NUM_MODEL_WORKERS to be 1?

src/sagemaker_xgboost_container/mms_patch/model_server.py
```diff
@@ -0,0 +1,178 @@
...
+DEFAULT_MMS_LOG_FILE = pkg_resources.resource_filename(sagemaker_inference.__name__, '/etc/log4j.properties')
+DEFAULT_MMS_MODEL_DIRECTORY = os.path.join(os.getcwd(), '.sagemaker/mms/models')
+DEFAULT_MMS_MODEL_NAME = 'model'
+
+PYTHON_PATH_ENV = 'PYTHONPATH'
+REQUIREMENTS_PATH = os.path.join(code_dir, "requirements.txt")
+MMS_NAMESPACE = "com.amazonaws.ml.mms.ModelServer"
+
+
+def start_model_server(is_multi_model=False, handler_service=DEFAULT_HANDLER_SERVICE, config_file=None):
```
ChuyangDeng: Is this going to be merged in sagemaker-inference-toolkit? 

If so, this change will require us to change framework containers code (to set is_multi_model value). I would let sagemaker-inference-toolkit to get the SAGEMAKER_MULTI_MODEL env var value and determine which mms config to use so that mxnet/pytorch container does not have to make code changes.

ericangelokim: I'm really not a fan of using env vars; you don't know what you might be overriding, and you don't know when it will be overriden. If the value is not going to be used in any other context it should be passed as a method argument. This file should be useable as it has a default value for the method arguments.

src/sagemaker_xgboost_container/mms_patch/model_server.py
```diff
@@ -0,0 +1,178 @@
...
+@retry(stop_max_delay=10 * 1000)
+def _retrieve_mms_server_process():
+    mms_server_processes = list()
+
+    for process in psutil.process_iter():
+        if MMS_NAMESPACE in process.cmdline():
+            mms_server_processes.append(process)
+
+    if not mms_server_processes:
+        raise Exception("mms model server was unsuccessfully started")
```
cbalioglu: Something is off with this message :)

ericangelokim: This was copied from mms_server code; I'm not really sure what this was supposed to catch either. I'll follow up with them during upstream merge

src/sagemaker_xgboost_container/serving.py
```diff
@@ -31,7 +30,7 @@
 USER_HANDLER_SERVICE = user_module_handler_service.__name__
 
 PORT = 8080
-MAX_PAYLOAD_IN_MB = 6
+DEFAULT_MAX_CONTENT_LEN = 6 * 1024 * 1024
```
aws-patlin: This commit looks good to me.

docker/0.90-2/resources/mms/ExecutionParameters.java
```diff
@@ -0,0 +1,86 @@
...
+        description = "Execution parameters endpoint")
+public class ExecutionParameters extends ModelServerEndpoint {
+
+    @Override
+    public void doGet(Request req, Response rsp, Context ctx) throws IOException {
+        Properties prop = ctx.getConfig();
+        SagemakerXgboostResponse r = new SagemakerXgboostResponse();
+        r.setMaxConcurrentTransforms(Integer.parseInt(prop.getProperty("NUM_WORKERS", "1")));
+        r.setBatchStrategy("MULTI_RECORD");
+        r.setMaxPayloadInMB(Integer.parseInt(prop.getProperty("SAGEMAKER_MAX_REQUEST_SIZE", "6")));
```
aws-patlin: SAGEMAKER_MAX_REQUEST_SIZE is a value in bytes. I think you need to divide by 1024*1024

ericangelokim: Yeap, that's what I was changing, and i'm having trouble building the changes right now.

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
```diff
@@ -0,0 +1,197 @@
...
+def _get_sparse_matrix_from_libsvm(payload):
+    pylist = map(lambda x: x.split(' '), payload.split('\n'))
+    colon = ':'
+    row = []
+    col = []
+    data = []
+    for row_idx, line in enumerate(pylist):
+        for item in line:
+            if colon in item:
+                col_idx = item.split(colon)[0]
```
iyerr3: Easier to say `col_idx, val = item.split(colon)`

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
```diff
@@ -0,0 +1,197 @@
...
+                val = item.split(colon)[1]
+                row.append(row_idx)
+                col.append(col_idx)
+                data.append(val)
+
+    row = np.array(row)
```
iyerr3: don't we need a `.astype(np.int)` here?

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
```diff
@@ -0,0 +1,197 @@
...
+    col = np.array(col).astype(np.int)
+    data = np.array(data).astype(np.float)
+    if not (len(row) == len(col) and len(col) == len(data)):
```
iyerr3: `len(row) == len(col) == len(data)` should work

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
```diff
@@ -0,0 +1,197 @@
...
+    return csr_matrix((data, (row, col)))
+
+
+def predict(booster, model_format, dtest, content_type):
+    if model_format == 'pkl_format':
+        x = len(booster.feature_names)
+        y = len(dtest.feature_names)
+
+        if content_type == 'text/x-libsvm' or content_type == 'text/libsvm':
+            if y > x + 1:
```
iyerr3: What's the context of `+1` here and in CSV? Is that because the labels could be included in the test data?

docker/0.90-2/final/Dockerfile.cpu
```diff
@@ -1,7 +1,52 @@
...
+########################
+# Install dependencies #
+########################
+COPY requirements.txt /requirements.txt
+RUN pip install -r /requirements.txt && rm /requirements.txt
+
+###########################
+# Copy wheel to container #
+###########################
+COPY dist/sagemaker_xgboost_container-1.0-py2.py3-none-any.whl /sagemaker_xgboost_container-1.0-py2.py3-none-any.whl
```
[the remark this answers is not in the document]
ericangelokim: Legacy. I can remove in this iteration if you think its important

docker/0.90-2/final/Dockerfile.cpu
```diff
@@ -1,7 +1,52 @@
...
+RUN pip install --no-cache /sagemaker_xgboost_container-1.0-py2.py3-none-any.whl && \
+    rm /sagemaker_xgboost_container-1.0-py2.py3-none-any.whl
+
+##############
+# DMLC PATCH #
+##############
+# TODO: remove after making contributions back to xgboost for tracker.py
```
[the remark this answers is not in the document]
ericangelokim: Amir and I discussed this and decided that you can create the patch from the source code so decided to keep the patch file as is in the repo.

Ideally we can merge this back when we have bandwidth.

docker/0.90-2/final/Dockerfile.cpu
```diff
@@ -1,7 +1,52 @@
...
+RUN mkdir -p /home/model-server/tmp
+RUN chown -R model-server /home/model-server
+
+# Copy MMS configs
+COPY docker/0.90-2/resources/mms/config.properties /home/model-server
+ENV XGBOOST_MMS_CONFIG=/home/model-server/config.properties
+
+# Copy execution parameters endpoint plugin for MMS
+RUN mkdir -p /tmp/plugins
+COPY docker/0.90-2/resources/mms/endpoints-1.0.jar /tmp/plugins
```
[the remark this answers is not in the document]
ericangelokim: Makes sense. Will add this and use to substitute in raw strings.

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
ericangelokim: Seems to pass previous container test behavior

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
ericangelokim: To be honest, I'm not sure. This was taken from the legacy xgboost model and I never questioned it because it was passing container tests. Might be worth doing a deep dive.

src/sagemaker_xgboost_container/serving.py
ericangelokim: Changing

test/unit/test_data_utils.py
```diff
@@ -160,9 +178,18 @@ def test_parse_protobuf_dmatrix_pipe(self):
...
                 reader = data_utils.get_recordio_protobuf_dmatrix
                 is_pipe = True
                 self._check_piped_dmatrix(pb_path, pipe_path, pipe_dir, reader, 5, 5, is_pipe)
+
+    def test_parse_sparse_protobuf_dmatrix(self):
+        pb_file_paths = ['sparse']
+
+        for file_path in pb_file_paths:
+            with self.subTest(file_path=file_path):
+                pb_path = os.path.join(self.resource_path, 'recordio_protobuf', file_path)
```
aws-patlin: Change `self.resource_path` to `self.data_path`, as per the other tests.

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
aws-patlin: I also thought it might have to do with the label being included in the actual data of the DMatrix, but I'm not sure if that really makes sense. Locally I was able to confirm that there is an error if the feature names do not match. So the `y == x + 1` case doesn't seem like it should work.

docker/0.90-1/resources/mms/ExecutionParameters.java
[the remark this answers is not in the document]
ericangelokim: Will do so.

aws-patlin:
Do you still need the `src/sagemaker_xgboost_container/algorithm_mode/serve.py` file?

ericangelokim:
No, that's a good point. I'll remove the serve.py file. and rerun container tests.

ericangelokim:
Yes. I actually looked into adding the pytest runs in the Dockerfile, but I didn't have time to test in Codebuild because they bumped image versions and I couldn't set up the project in my repo. This is a quick follow up that needs to be done; Rizwan already took a look into this as well.

test-requirements.txt
```diff
@@ -5,9 +5,15 @@ coverage
 docker-compose
 flake8
 mock
+mxnet-model-server
 pytest
 pytest-cov
 pytest-xdist
+python-dateutil==2.8.0
```
aws-patlin: When I tried building my inference changes on top of this branch, I noticed that during the installation of the final Dockerfile.cpu, there's an error regarding the python-dateutil version conflicting with botocore when installing requirements.txt. Since this also affects the container build, can you move this to requirements.txt instead of having it in here?

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
aws-patlin: I tried running container tests after removing these checks, and it actually breaks the inference tests. Not sure why this works, but looks like it is necessary.

docker/0.90-2/resources/mms/ExecutionParameters.java
```diff
@@ -0,0 +1,96 @@
...
+                .write(
+                        new GsonBuilder()
+                                .setPrettyPrinting()
+                                .create()
+                                .toJson(r)
+                                .getBytes(StandardCharsets.UTF_8));
+    }
+
+    /** Response for Model server endpoint */
+    public static class SagemakerXgboostResponse {
```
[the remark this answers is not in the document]
ericangelokim: I wanted to distinguish this from the example code in the mxnet model server repo, hence the more specific name.

src/sagemaker_xgboost_container/algorithm_mode/handler_service.py
```diff
@@ -0,0 +1,197 @@
...
+                           validate_features=False)
+
+
+class HandlerService(DefaultHandlerService):
+    """Handler service that is executed by the model server.
+    Determines specific default inference handlers to use based on the type MXNet model being used.
+    This class extends ``DefaultHandlerService``, which define the following:
+        - The ``handle`` method is invoked for all incoming inference requests to the model server.
+        - The ``initialize`` method is invoked at model server start up.
+    Based on: https://github.com/awslabs/mxnet-model-server/blob/master/docs/custom_service.md
```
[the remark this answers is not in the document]
ericangelokim: Good catch, I'm surprised I put the master branch here myself. Will fix

PR #44, opened 2019-11-07
Mms user mme

ericangelokim:
*Issue #, if available:*

*Description of changes:*
Enabled MME support for user script mode. Note that this only uses one user module.

Tested:
Notebook tests
Tox
Repository integ tests pass

Note that boto does not have MME API calls available yet; we will need to think about how we can set up tests in pipelines for MME support changes.


ericangelokim:
Reviewed with Julio, who gave verbal approval for the changes.

PR #45, opened 2019-11-08
Use Python 3.7 installed with conda

edwardjkim:
*Issue #, if available:*

*Description of changes:*

**R2**

Removed unnecessary pip download and installation. pip is available in Conda.

*Tests*
```
pytest test/integration
```
Internal container tests and integration tests all passed (except HPO).

**R1**

Since we now use conda to install ml-io, this PR proposes moving to Python 3.7. Previously `apt-get install python3` installed ubuntu 16's default python version 3.5. By removing `apt-get install python3`, the conda-installed Python is used.

*Tests:*
```
pytest test/integration
```
Internal functional tests and integration tests all passed (except HPO).

### DIFF
--- a/docker/0.90-2/base/Dockerfile.cpu
+++ b/docker/0.90-2/base/Dockerfile.cpu
@@ -24,12 +24,6 @@ RUN conda update -y conda && \
     conda install -c conda-forge pyarrow=0.14.1 && \
     conda install -c mlio -c conda-forge mlio-py=0.1
 
-# Install pip
-RUN cd /tmp && \
-    curl -O https://bootstrap.pypa.io/get-pip.py && \
-    python3 get-pip.py && \
-    rm get-pip.py
-
 # Python won’t try to write .pyc or .pyo files on the import of source modules
 # Force stdin, stdout and stderr to be totally unbuffered. Good for logging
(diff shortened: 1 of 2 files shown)


aws-patlin:
LGTM. Are there any concerns with raising the python version?

docker/0.90-2/base/Dockerfile.cpu
```diff
@@ -2,29 +2,38 @@ FROM ubuntu:16.04
...
-RUN conda update -y conda
-RUN conda install -c conda-forge pyarrow=0.14.1
-RUN conda install -c mlio -c conda-forge mlio-py=0.1
+
+RUN conda update -y conda && \
+    conda install -c conda-forge pyarrow=0.14.1 && \
+    conda install -c mlio -c conda-forge mlio-py=0.1
+
+# Install pip
+RUN cd /tmp && \
```
cbalioglu: pip should be available in the default Conda environment. Is there a specific reason why we are explicitly installing pip?

edwardjkim: Right, this was a left over from the previous version. Let me remove it.

edwardjkim:
Closing. Will raise another one.

PR #46, opened 2019-11-11
Add subsampling support for CSV and RecordIO-protobuf datasets

cbalioglu:
This PR introduces support for subsampling-on-read to CSV datasets read from the SageMaker pipe mode. What subsampling means in this context is reducing the size of dataset artificially to ensure that it fits into the memory of a given host machine.

The logic implemented here simply takes the first `subsample_ratio * mini_batch_size` records of the mini-batch and discards the rest; this effectively reduces the size of the dataset by the specified ratio.

Also included in this PR is an upgrade to ML-IO v0.2. The new version of ML-IO now uses Apache Arrow v0.15.1 which brings lots of bug fixes and performance improvements.

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
@@ -202,5 +202,11 @@ def initialize(metrics):
                                              required=False),
         hpv.IntegerHyperparameter(name="seed", range=hpv.Interval(min_open=-2**31, max_open=2**31-1),
                                   required=False),
+
+        # For internal use only!
+        hpv.ContinuousHyperparameter(name="_subsample_ratio_on_read",
+                                     range=hpv.Interval(min_open=0,
+                                                        max_open=1),
+                                     required=False),
         )
     return hyperparameters
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -31,7 +31,12 @@ from sagemaker_xgboost_container.constants.xgb_constants import CUSTOMER_ERRORS
 logger = logging.getLogger(__name__)
 
 
-def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights=0, is_pipe=False):
+def get_validated_dmatrices(train_path,
+                            validate_path,
+                            content_type,
+                            csv_weights=0,
+                            is_pipe=False,
+                            subsample_ratio_on_read=None):
     """Get training and validation Data Matrices for XGBoost training.
 
     Check size and format of both training and validation data channels, and return parsed
@@ -39,9 +44,11 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
 
     :param train_path:
     :param validate_path:
-    :param content_type: Content type of data. Supports 'libsvm' or 'csv'
+    :param content_type: Content type of data. Supports 'libsvm', 'csv', 'parquet', and 'recordio-protobuf'.
     :param csv_weights: 1 if instance weights are in the second column of csv data files; otherwise, 0
     :param is_pipe: Boolean to indicate if data is being read in pipe mode
+    :param subsample_ratio_on_read: None or a value in (0, 1) to indicate how much of the dataset should
+            be read into memory.
     :return: Parsed xgb.DMatrix
     """
     train_files_size = get_size(train_path, is_pipe) if train_path else 0
@@ -56,7 +63,11 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
         if val_files_size > 0:
             validate_data_file_path(validate_path, content_type)
 
-    train_dmatrix = get_dmatrix(train_path, content_type, csv_weights=csv_weights, is_pipe=is_pipe) \
+    train_dmatrix = get_dmatrix(train_path,
+                                content_type,
+                                csv_weights=csv_weights,
+                                is_pipe=is_pipe,
+                                subsample_ratio_on_read=subsample_ratio_on_read) \
         if train_files_size > 0 else None
     val_dmatrix = get_dmatrix(validate_path, content_type, is_pipe=is_pipe) \
         if val_files_size > 0 else None
@@ -99,9 +110,15 @@ def sagemaker_train(train_config, data_config, train_path, val_path, model_dir, 
     input_mode = validated_data_config['train'].get("TrainingInputMode")
     csv_weights = validated_train_config.get("csv_weights", 0)
     is_pipe = (input_mode == Channel.PIPE_MODE)
+    subsample_ratio_on_read = validated_train_config.get("_subsample_ratio_on_read", None)
 
     validation_channel = validated_data_config.get('validation', None)
-    train_dmatrix, val_dmatrix = get_validated_dmatrices(train_path, val_path, file_type, csv_weights, is_pipe)
+    train_dmatrix, val_dmatrix = get_validated_dmatrices(train_path,
+                                                         val_path,
+                                                         file_type,
+                                                         csv_weights,
+                                                         is_pipe,
+                                                         subsample_ratio_on_read)
 
     checkpoint_dir = checkpoint_config.get("LocalPath", None)
 
--- a/src/sagemaker_xgboost_container/data_utils.py
+++ b/src/sagemaker_xgboost_container/data_utils.py
@@ -30,7 +30,7 @@ from sagemaker_algorithm_toolkit import exceptions as exc
 from sagemaker_xgboost_container.constants import xgb_content_types
 
 
-BATCH_SIZE = 4000
+BATCH_SIZE = 1000
 
 CSV = 'csv'
 LIBSVM = 'libsvm'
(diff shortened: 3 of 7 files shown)


src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -300,15 +300,17 @@ def _get_csv_dmatrix_file_mode(files_path, csv_weights):
     return dmatrix
 
 
-def _get_csv_dmatrix_pipe_mode(pipe_path, csv_weights):
+def _get_csv_dmatrix_pipe_mode(pipe_path, csv_weights, subsample_on_read):
```
iyerr3: 1. Should we rename to indicate a ratio? I read this right now as a boolean. 
2. Should it have a default of `1` i.e. no subsampling?

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -42,6 +47,8 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
     :param content_type: Content type of data. Supports 'libsvm' or 'csv'
```
aws-patlin: I noticed I didn't update this comment to include parquet and recordio-protobuf. Can you quickly fix this?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -300,15 +300,17 @@ def _get_csv_dmatrix_file_mode(files_path, csv_weights):
...
 
     :param pipe_path: SageMaker pipe path where CSV formatted training data is piped
     :param csv_weights: 1 if instance weights are in second column of CSV data; else 0
+    :param subsample_on_read: None or a value in (0, 1] to indicate how much of the dataset should
+            be read into memory.
     :return: xgb.DMatrix or None
     """
     try:
-        dataset = [mlio.SageMakerPipe(pipe_path)]
+        dataset = [mlio.SageMakerPipe(pipe_path, fifo_id=0)]
```
aws-patlin: Does the fifo_id not default to 0?

cbalioglu: Yes, it does. The reason I had to explicitly specify it is because we have unit tests for pipe mode that all run in the same process. Without this parameter secondary unit tests fail because they look for FIFO 1, but the pipe mode emulator gets restarted with each unit test. Ideally those tests should not be unit but integration tests in the future. Explicitly specifying a `fifo_id` of `0` does not change the behavior, but it is superfluous

src/sagemaker_xgboost_container/algorithm_mode/train.py
cbalioglu: Sure, will do.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -505,33 +517,45 @@ def get_dmatrix(data_path, content_type, csv_weights=0, is_pipe=False):
...
+
+    if os.path.isfile(data_path) or is_pipe:
+        files_path = data_path
+    elif not is_pipe:
+        for root, dirs, files in os.walk(data_path):
+            if dirs == []:
+                files_path = root
+                break
+    if content_type.lower() == CSV:
+        dmatrix = get_csv_dmatrix(files_path,
```
iyerr3: There is value in having this subsampling for all formats. Is it possible to abstract the code to ensure it applies for all reads?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -505,33 +517,45 @@ def get_dmatrix(data_path, content_type, csv_weights=0, is_pipe=False):
...
+                                  csv_weights,
+                                  is_pipe,
+                                  subsample_on_read)
+    elif content_type.lower() == LIBSVM:
+        dmatrix = get_libsvm_dmatrix(files_path, is_pipe)
+    elif content_type.lower() == PARQUET:
+        dmatrix = get_parquet_dmatrix(files_path, is_pipe)
+    elif content_type.lower() == RECORDIO_PROTOBUF:
+        dmatrix = get_recordio_protobuf_dmatrix(files_path, is_pipe)
```
ericangelokim: Can we add exception to throw back to customer if no support content type is selected? Might as well make this defensive on its own

src/sagemaker_xgboost_container/data_utils.py
ericangelokim: Agree with the first point.

The second point is debateable because if the value is None then it the code flow doesn't even go through the subsampling section. However, I'm also not fond of overloading arguments (ie, None in one case, and then ratio in another. Is this pythonic best practices?).

We can also have default be 1, and the check for subsampling does:

```
if submsampling is < 1 and subsampling > 0:
    do stuff
```

And this way we don't need a None default.

cbalioglu: 1. Yeah I can rename it to `subsample_ratio_on_read`.
2. This is an internal function. The default value (`None`) is set in `get_csv_dmatrix`.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -318,14 +320,18 @@ def _get_csv_dmatrix_pipe_mode(pipe_path, csv_weights):
...
-                tmp = [as_numpy(feature).squeeze() for feature in example]
-                tmp = np.array(tmp)
-                if len(tmp.shape) > 1:
-                    # Columns are written as rows, needs to be transposed
-                    tmp = tmp.T
-                else:
+                tmp = np.column_stack([as_numpy(f) for f in example])
+
+                # If a subsample ratio is specified, only take the first
+                # `BATCH_SIZE * subsample_on_read` records and discard the rest
```
aws-patlin: I thought the intention was to perform some systematic sampling (i.e. take every nth row) instead of taking the first group of records.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -318,14 +320,18 @@ def _get_csv_dmatrix_pipe_mode(pipe_path, csv_weights):
...
+                # of the mini-batch.
+                if subsample_on_read and subsample_on_read < 1.0:
+                    tmp = tmp[:max(int(tmp.shape[0] * subsample_on_read), 1)]
+
+                if len(tmp.shape) == 1:
```
aws-patlin: Can the output of column_stack be a 1-D array? If not, then this check isn't necessary anymore.

src/sagemaker_xgboost_container/data_utils.py
cbalioglu: If you think of `subsample_ratio_on_read` as an optional parameter, having a value of `None` vs. `1` has a semantic difference. As an API user with `None` you explicitly specify that you do not want to have any subsampling; with `1`, although the result will be the same as if `None` is specified, you do not know for sure whether there will be some kind of additional logic executed related to subsampling.

src/sagemaker_xgboost_container/data_utils.py
cbalioglu: After having further discussions this approach found to be more flexible.

src/sagemaker_xgboost_container/data_utils.py
iyerr3: FYI: XGBoost has a similar parameter (https://xgboost.readthedocs.io/en/latest/parameter.html#parameters-for-tree-booster) that has default value `1`. 

Q: Under what circumstance would an API user specify `subsample_ratio_on_read=None`? 
- If no subsampling is desired then user could just not specify the parameter. 
- Additionally if `subsample_ratio_on_read=1` then behavior *should* be same as `None`. 

So it's not clear to me when the `None` parameter should be used.

src/sagemaker_xgboost_container/data_utils.py
cbalioglu: I moved the subsampling logic into ML-IO. It means the logic is now available for all code paths that use ML-IO for data reading (as of today CSV/pipe and RecordIO-protobuf/[file, pipe])

src/sagemaker_xgboost_container/data_utils.py
cbalioglu: Added it as a last `else` clause.

src/sagemaker_xgboost_container/data_utils.py
aws-patlin: How does the subsampling interact with the batch size? I'm assuming the size of the example will still be equal to the batch size, but the reader will throw away however many records it reads based on the subsampling ratio until it reaches the batch size.

aws-patlin:
Did you remove the unit test you initially wrote?

cbalioglu:
> Did you remove the unit test you initially wrote?

Yes, ML-IO uses a different logic than what I had here before. If the batch size is less than or equal to the (remaining) dataset size, it will always get filled in. What ML-IO does is, after reading a mini-batch, it skips `(batch_size / subsample_ratio) - batch_size` records and starts reading the next mini-batch from there. Practically it does the same thing as what was previously implemented here only with the difference that it discards records before putting them into a mini-batch.

src/sagemaker_xgboost_container/data_utils.py
cbalioglu: I changed the accepted range of values to (0, 1); meaning, if specified, it has to be greater than zero and less than 1. If `None`, then no subsampling is performed. I still think that `None` is more distinct than `1` to indicate that no subsampling should be performed.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -505,33 +509,56 @@ def get_dmatrix(data_path, content_type, csv_weights=0, is_pipe=False):
...
+        dmatrix = get_libsvm_dmatrix(files_path, is_pipe)
+    elif content_type.lower() == PARQUET:
+        if subsample_ratio_on_read:
+            logging.warning(
+                "The subsample_ratio_on_read hyperparameter is not used for "
+                "Parquet datasets.")
+
+        dmatrix = get_parquet_dmatrix(files_path, is_pipe)
+    elif content_type.lower() == RECORDIO_PROTOBUF:
+        dmatrix = get_recordio_protobuf_dmatrix(files_path, is_pipe)
```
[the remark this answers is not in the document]
cbalioglu: Yes, my bad. Fixed it.

src/sagemaker_xgboost_container/data_utils.py
cbalioglu: Yep, deleted it in the second revision.

cbalioglu:
> For eg. dataset = 300, batch_size = 100, subsample_ratio = 0.5, expected output = 150
> So by above logic, we'll read 100 rows, then skip 100 rows (100/0.5 - 100) and then read 100 again. So we'll end up with 200 rows instead of 150. Am I reading this wrong?

If your dataset is fairly small, as in your example, then yes, it does not match the exact ratio. I explicitly have a remark in the API documentation that mentions this fact. ML-IO has no other way to subsample a dataset. If you do not know the size of the dataset in advance and the user requests a mini-batch size of 100, your only option is to skip `n` number of records in between the mini-batches. If your dataset is sufficiently large, then the actual ratio will converge to the requested ratio.

aws-patlin:
You need to update tox.ini to use the updated mlio and pyarrow versions.

test/unit/test_data_utils.py
```diff
@@ -14,8 +14,8 @@
 import unittest
 import os
 from pathlib import Path
+import shutil
 import signal
```
aws-patlin: You can remove this, since you also removed the only instance it was used.

src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
```diff
@@ -202,5 +202,8 @@ def eval_metric_dep_validator(value, dependencies):
                                              required=False),
         hpv.IntegerHyperparameter(name="seed", range=hpv.Interval(min_open=-2**31, max_open=2**31-1),
                                   required=False),
+
+        # For internal use only!
+        hpv.ContinuousHyperparameter(name="_subsample_ratio_on_read", range=hpv.Interval(min_open=0, max_open=1), required=False),
```
aws-patlin: This line is too long. Can you break it up, one arg per line?

cbalioglu: I broke it up. Overall the whole code base needs to be linted though. Looks like no one has run `flake8` or a similar tool yet.

test/unit/test_data_utils.py
cbalioglu: It is still being used. I deleted the `_clear_folder` function and started using `shutil.rmtree` in `_check_piped_dmatrix`.

rizwangilani:
I was testing this, and seems like we do not support 1 as value for subsample_ratio_on_read hyperparameter. Is this the best experience for the customer?

cbalioglu:
> I was testing this, and seems like we do not support 1 as value for subsample_ratio_on_read hyperparameter. Is this the best experience for the customer?

First a sidenote: `_subsample_ratio_on_read` is an internal "hyperparameter", so it should not be used by customers.

Now back to the main part of the question: My intuition was that if subsampling is requested, it has to be in the range of (0, 1). Subsampling with 1 does not make any sense as you are practically not subsamling anything in such case. So if you specify `None` (which is the default value), subsampling is disabled; if you specify a value it has to be a floating-point number greater than 0 and less than 1. I personally find a value of `None` more descriptive than `1` if subsampling is not desired. With `1` it is not clear whether we are actually performing subsampling.

PR #49, opened 2019-11-13
Dynamically set heap and direct memory limits.

aws-patlin:
*Description of changes:*
The `MaxDirectMemorySize` parameter limits the amount of direct memory that the JVM can use. This has a direct effect of limiting the size of inference requests we can make, causing `OutOfDirectMemoryError`s as soon as the 10MB limit is reached. This was temporarily raised during testing to 1GB. 

After raising the direct memory limit, the `Xmx` parameter, which limits the max heap size, began causing `OutOfMemoryError`s when testing batch transform jobs. After raising that limit as well, the errors were resolved.

The `Xmx` and `MaxDirectMemorySize` vmargs are now set based on the number of workers * the max payload size * a buffering factor of 1.2  + a base amount of 128mb, which is theoretically sufficient for the server when under full load (all workers occupied). Environment variables are not parsed for vmargs (MMS has not implemented this yet), so the current workaround is to write the values to the config.properties file before model startup.

Change were tested by running container tests and batch transform jobs with small (<5MB) and larger payloads (up to 20MB).

### DIFF
--- a/src/sagemaker_xgboost_container/serving.py
+++ b/src/sagemaker_xgboost_container/serving.py
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 from __future__ import absolute_import
 import logging
+from math import ceil
 import multiprocessing
 import os
 import subprocess
@@ -30,6 +31,7 @@ USER_HANDLER_SERVICE = user_module_handler_service.__name__
 
 PORT = 8080
 DEFAULT_MAX_CONTENT_LEN = 6 * 1024**2
+MAX_CONTENT_LEN_LIMIT = 20 * 1024**2
 
 
 def _retry_if_error(exception):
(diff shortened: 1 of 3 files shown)


docker/0.90-2/resources/mms/config.properties
```diff
@@ -1,4 +1,4 @@
-vmargs=-Xmx128m -XX:-UseLargePages -XX:+UseG1GC -XX:MaxMetaspaceSize=32M -XX:MaxDirectMemorySize=10m -XX:+ExitOnOutOfMemoryError
+vmargs=-Xmx128m -XX:-UseLargePages -XX:+UseG1GC -XX:MaxMetaspaceSize=32M -XX:MaxDirectMemorySize=1G -XX:+ExitOnOutOfMemoryError
```
cbalioglu: Have you done any measurement? Why 1G and, say, not 100M?

ericangelokim: 1GB came from a quick napkin calculation. The buffer is shared across the JVM, and if each workers was concurrently receiving 20MB payloads that would allow roughly 50 workers to work in parallel. The largest number of CPU's a sagemaker instance has is 96 based on https://aws.amazon.com/sagemaker/pricing/instance-types/ 

This is obviously not a perfect solution. I think if we had bandwidth i would want to test calculating num_cpu * max_payload_size and setting that as a env var that can be parsed by MMS to use in the config file. This change is just to unblock Flintstone to use 20MB payloads

ericangelokim:
Please leave a comment explaining the TODO that should follow up

iyerr3:
Maybe this was discussed offline: is the parameter mandatory? Can MMS provide the functionality of detecting the right value per available memory?

aws-patlin:
> Maybe this was discussed offline: is the parameter mandatory? Can MMS provide the functionality of detecting the right value per available memory?

I think it becomes harder to control the memory behavior once we start hosting multiple containers on an endpoint, so it may be better to have a well defined limit. It also seems like removing the `MaxDirectMemorySize` parameter falls back to some other value, as I was seeing the same error around 100MB (could be tied to the `Xmx` parameter, which we also needed to increase).

aws-patlin:
To confirm the default behavior when `Xmx` and `MaxDirectMemorySize` are not set:
* Max heap size defaults to ¼ of the physical memory, up to 1GB. [Ref](https://stackoverflow.com/questions/28272923/default-xmxsize-in-java-8)
* Max direct memory defaults to the value of the max heap size. [Ref](https://stackoverflow.com/questions/3773775/default-for-xxmaxdirectmemorysize)

In this case, it makes sense for us to be setting the limits to higher values depending on the instance type instead of relying on these defaults.

src/sagemaker_xgboost_container/serving.py
```diff
@@ -65,22 +66,42 @@ def _set_mms_configs(is_multi_model, handler):
...
+    # following code.
+    try:
+        with open('/home/model-server/config.properties.tmp', 'r') as f:
+            with open('/home/model-server/config.properties', 'w+') as g:
+                g.write("vmargs=-XX:-UseLargePages -XX:+UseG1GC -XX:MaxMetaspaceSize=32M -XX:+ExitOnOutOfMemoryError "
+                        + "-Xmx" + os.environ["SAGEMAKER_MAX_HEAP_SIZE"]
+                        + " -XX:MaxDirectMemorySize=" + os.environ["SAGEMAKER_MAX_DIRECT_MEMORY_SIZE"] + "\n")
+                g.write(f.read())
+    except Exception:
+        pass
```
edwardjkim: Is the server going to start without `config.properties`? If not, is it ok to fail silently here?

aws-patlin: I added this try-except because the path to the config.properties only exists on the actual container, which was causing the unit tests to fail. Is there a better way to do this?

edwardjkim: I just read the code comments above the try-except block. If the plan is to remove this code in the near future, then I guess it's okay. Reading `try: something, except Exception: pass` made me think "it's okay to fail this block".

aws-patlin: Yeah, ideally we want to have the config.properties as it was before and use the environment variables to get the vmargs. The issue is that the MMS team hasn't implemented environment variable parsing for vmargs, so this is a temporary solution until they release that feature.

ericangelokim: Minor: instead of concat'ing strings, can you use format and use variable names to put values in the string for readability?

Also, I'm ok with this temporary solution but another solution worth thinking about would've been to take files and a dict and do overrides to create a new config file and its location.

PR #52, opened 2019-11-21
Increase response size.

ericangelokim:
*Description of changes:*
Increased respionse size max so that it matches payload size (extreme case, if user uses one feature model and the output is same size as payload.

*Test*
Configured in local run.
```
Maximum Response Size: 20971520
Maximum Request Size: 20971520
```


rizwangilani:
do we still need to merge this?

PR #58, opened 2019-11-23
Use Python 3.7 installed with conda

edwardjkim:
*Description of changes:*

Same change as #45 but against `0.90-1`.


aws-patlin:
Do we really want to make this change to the 0.90-1 image? I thought we were concerned about making a potentially backwards incompatible change with the python version bump.

edwardjkim:
No, we don't. Sorry, please ignore.

PR #65, opened 2019-11-25
Add support to parse MAX_CONTENT_LENGTH during inference.

ericangelokim:
*Description of changes:*

Increase max content length up to 20MB based one user defined environment variable/

*Tests run*
```tox```
Containers tests and integration tests pass

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/serve.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/serve.py
@@ -35,6 +35,19 @@ SAGEMAKER_BATCH = os.getenv("SAGEMAKER_BATCH")
 logging = integration.setup_main_logger(__name__)
 
 
+def _get_max_content_size():
+    max_payload_size = 20 * 1024 ** 2
+    # NOTE: 6 MB max content length = 6 * 1024 ** 2
+    content_len = int(os.getenv("MAX_CONTENT_LENGTH", '6291456'))
+    if content_len < max_payload_size:
+        return content_len
+    else:
+        return max_payload_size
+
+
+PARSED_MAX_CONTENT_LENGTH = _get_max_content_size()
+
+
 def number_of_workers():
     return multiprocessing.cpu_count()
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/sagemaker_xgboost_container/algorithm_mode/serve.py [hunk 1]
```diff
@@ -35,6 +35,19 @@
 logging = integration.setup_main_logger(__name__)
 
 
+def _get_max_content_size():
```
aws-patlin: Is there a reason this uses the word "size" when everything else refers to "length"?

src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -57,11 +70,8 @@ def load(self):
 
 class ScoringService(object):
     PORT = os.getenv("SAGEMAKER_BIND_TO_PORT", 8080)
-    # NOTE: 6 MB max content length
-    MAX_CONTENT_LENGTH_MB = int(os.getenv("MAX_CONTENT_LENGTH", '6'))
-    MAX_CONTENT_LENGTH = MAX_CONTENT_LENGTH_MB * 1024 ** 2
-
     MODEL_PATH = os.getenv(sm_env_constants.SM_MODEL_DIR)
+    MAX_CONTENT_LENGTH = PARSED_MAX_CONTENT_LENGTH
```
aws-patlin: Is there any reason to have both `MAX_CONTENT_LENGTH` and `PARSED_MAX_CONTENT_LENGTH`? It looks like the only time you use `PARSED_MAX_CONTENT_LENGTH` is in line 179, when setting the value of `MaxPayloadInMB`. I would get rid of `PARSED_MAX_CONTENT_LENGTH` and set `MAX_CONTENT_LENGTH` directly to the value of `_get_max_content_size()`.

ericangelokim: Settled offline, ScoringService has MAX_CONTENT_LENGTH and also would like to avoid parsing env vars more than once.

PR #69, opened 2019-11-26
Fix incorrect keyword argument in DMatrix creation

Jarryd-rk:
*Issue #, if available:*

*Description of changes:*
The instance weight parameter is incorrectly called weights in the DMatrix creation. As a result, if customers specify csv_weights=1 the container fails with algorithm error TypeError: __init__() got an unexpected keyword argument 'weights'

https://xgboost.readthedocs.io/en/release_0.90/python/python_api.html#xgboost.DMatrix


ericangelokim:
Hello, thank you for this catch. Can you include a unit-test that verifies this change?

ericangelokim:
Replacing with PR that has unit tests: https://github.com/aws/sagemaker-xgboost-container/pull/79

PR #71, opened 2019-11-26
Use Python 3.6 installed with conda

edwardjkim:
*Issue #, if available:*

*Description of changes:*

Similar to #45 but I think upgrading to Python 3.6 is a better solution because:

1. Came across [this issue](https://github.com/apache/incubator-superset/issues/8349) and it looks like Gunicorn is recommending 3.6 because they test against 3.6,
2. SageMaker uses 3.6 currently.

Tests: flake8, pytest test/unit, internal functional tests, all internal integration tests pass except HPO.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ setup(
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python",
         'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
     ],
 
     install_requires=read("requirements.txt"),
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -29,6 +29,7 @@ def read(fname):
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python",
         'Programming Language :: Python :: 3.5',
```
aws-patlin: Do we want to keep this line for this version of the container? Will it still run with Python 3.5?

edwardjkim: Yes, I think we want to keep 3.5 for this commit because at least the SXC module will work with 3.5. But we will have to drop 3.5 in the next commit when we install tornasole.

aws-patlin: Ah, got it.

PR #78, opened 2019-12-02
Fix script mode training hang with logging enabled.

aws-patlin:
*Description of changes:*
A user discovered that after a recent release, their training script stops to work when adding a stdout stream handler to the logger. The issue was traced back to this commit in sagemaker-containers.

capture_error=True appends stderr to the error message that gets thrown if training fails. For context, this was specifically a workaround for PyTorch, which can throw a specific error even if training succeeds, so I don't believe this is necessary for XGBoost.

Corresponding PR for 0.90-1: https://github.com/aws/sagemaker-xgboost-container/pull/77

Testing:
Using the prod image with capture_error enabled, the training script would hang with no log output. With capture_error disabled on a custom image, I was able to complete the training job successfully with the expected log output.

### DIFF
--- a/src/sagemaker_xgboost_container/training.py
+++ b/src/sagemaker_xgboost_container/training.py
@@ -82,7 +82,7 @@ def train(training_environment):
         logger.info('Invoking user training script.')
         framework.modules.run_module(training_environment.module_dir, training_environment.to_cmd_args(),
                                      training_environment.to_env_vars(), training_environment.module_name,
-                                     capture_error=True)
+                                     capture_error=False)
     else:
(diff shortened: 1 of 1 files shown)


ericangelokim:
Looks fine to me, can we get client-libraries to formally approve this PR too?

ericangelokim:
I think you copied the same change in this PR: https://github.com/aws/sagemaker-xgboost-container/pull/77

aws-patlin:
> I think you copied the same change in this PR: #77

This PR is to merge into 0.90-2, #77 is for master/0.90-1

iyerr3:
We don't need this if we're merging to master. This branch will be merged to master in a few days.

PR #82, opened 2019-12-03
Add missing eval_metrics to hp validation.

ericangelokim:
*Issue #, if available:*

*Test*
```tox```
Adding container tests in pipelines

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/metrics.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/metrics.py
@@ -12,43 +12,15 @@
 # language governing permissions and limitations under the License.
 from sagemaker_algorithm_toolkit import metrics as m
 
+from sagemaker_xgboost_container.constants.xgb_constants import XGB_METRIC_NAMES
+
 
 # https://docs.aws.amazon.com/sagemaker/latest/dg/xgboost-tuning.html
 def initialize():
-    return m.Metrics(
-        m.Metric(name="validation:accuracy",
-                 direction=m.Metric.MAXIMIZE,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/sagemaker_xgboost_container/algorithm_mode/metrics.py [hunk 1]
```diff
@@ -12,43 +12,15 @@
...
-                 regex=".*\\[[0-9]+\\].*#011validation-mse:(\\S+)"),
-        m.Metric(name="validation:ndcg",
-                 direction=m.Metric.MAXIMIZE,
-                 regex=".*\\[[0-9]+\\].*#011validation-ndcg:(\\S+)"),
-        m.Metric(name="validation:rmse",
-                 direction=m.Metric.MINIMIZE,
-                 regex=".*\\[[0-9]+\\].*#011validation-rmse:(\\S+)"))
+
+    metrics = [m.Metric(name="validation:{}".format(metric_name),
+                        direction=m.Metric.MAXIMIZE,
```
iyerr3: Not clear to me on how direction is being set based on metric name

iyerr3: We might have to split into two lists or have a tuple for each entry with second item indicating the direction.

ericangelokim: I've added it back as it was before, with the addition of aucpr.

iyerr3:
I liked your previous change to simplify the code. Was thinking we could split to `XGB_MAXIMIZE_METRICS` and `XGB_MINIMIZE_METRICS` to get the split behavior.

ericangelokim:
I would prefer we have one source too. I think have those two lists makes sense.

PR #102, opened 2020-03-27
Bump psutil and sagemaker-containers version

rizwangilani:
Bump psutil from 5.4.8 to 5.6.7 and sagemaker-containers from 2.5.6 to 2.8.3

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,11 +3,11 @@ gunicorn<20.0.0
 matplotlib
 numpy
 pandas>=0.24.0
-psutil==5.4.8  # sagemaker-containers requires psutil 5.4.8
+psutil==5.6.7  # sagemaker-containers requires psutil 5.6.7
 python-dateutil==2.8.0
 requests<2.21
 retrying==1.3.3
-sagemaker-containers>=2.5.6
+sagemaker-containers>=2.8.3
 scikit-learn
 scipy==1.2.2
 smdebug==0.4.13


requirements.txt
```diff
@@ -3,11 +3,11 @@ gunicorn<20.0.0
 matplotlib
 numpy
 pandas>=0.24.0
-psutil==5.4.8  # sagemaker-containers requires psutil 5.4.8
+psutil==5.6.6  # sagemaker-containers requires psutil 5.4.8
```
edwardjkim: Is the comment `sagemaker-containers requires psutil 5.4.8` still true?

rizwangilani: actually, sagemaker-containers is using psutil `5.6.7` in their` 2.8.3` version. updated PR.

edwardjkim:
- Did we run integration tests?
- Does sagemaker 2.8.3 have the fix for the requirements.txt issue? Is there a way for us to test this?
- Does XGBoost have network isolation?

aws-patlin:
@edwardjkim Just an FYI, the issues with requirements.txt and network isolation should not affect XGBoost. Those bugs were present in the code path of `modules.download_and_install` and `entrypoint.run`, while in XGBoost we call `modules.run_module`.

edwardjkim:
Thanks @aws-patlin, that makes sense. Can we just make sure to manually run integration tests as a sanity check before merging since this involves sagemaker-containers?

PR #105, opened 2020-04-27
Merge master into 0.90-2

aws-patlin:
Merge master into 0.90-2 to prepare for 1.0 release.

Merge conflicts fixed:
* requirements.txt modified
* ci/buildspec.yml removed

Testing:
Tox run successfully.


iyerr3:
I'm confused on why this is a PR? Shouldn't this be just a new branch creation.

aws-patlin:
> I'm confused on why this is a PR? Shouldn't this be just a new branch creation.

We already have a 0.90-2 branch. I figured it would be better to merge master into that branch.

iyerr3:
I think delete and create would be better as the `90-2` needs to be identical to master, isn't it?

aws-patlin:
> I think delete and create would be better as the `90-2` needs to be identical to master, isn't it?

Ok, sounds good.

PR #109, opened 2020-05-19
Support hyperparameter aliases

salmankhurshid1:
*Issue #, if available:*

*Description of changes:*

- Adds support for following hyperparameters and aliases:

   - **_eta -> learning_rate_**
   - **_gamma -> min_split_loss_**
   - **_lambda -> reg_lambda_**
   - **_alpha -> reg_alpha_**

Tested with `tox` and integration tests.


src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
```diff
@@ -244,15 +244,26 @@ def _visit(name, visited, stack):
...
+        hp_aliases = {"learning_rate": "eta", "min_split_loss": "gamma", "reg_lambda": "lambda", "reg_alpha": "alpha"}
+
+        for hp, value in user_hyperparameters.items():
+            if hp in hp_aliases:
+                tmp_user_hyperparameters[hp_aliases[hp]] = value
+            else:
+                tmp_user_hyperparameters[hp] = value
+        user_hyperparameters = tmp_user_hyperparameters
+
+        # NOTE: 1. Validate required or fill in default.
```
rizwangilani, line 258: This approach seems to handle aliases correctly. As an alternative, how about creating a method such as `_declare_alias(key_name, alias_name)` and use that to declare aliases when initializing hyperparametes? It may help with readability since aliases will be declared closer to where hps are initialized (as opposed to maintaining a list within the `validate` method itself). Similarly, moving logic to `_declare_alias()` will help with keeping code modular. 

What do you think?

rizwangilani, line 258: This approach seems to handle aliases correctly. As an alternative, how about creating a method such as `_declare_alias(key_name, alias_name)` within the `Hyperparameters` class and use that to declare aliases when initializing hyperparametes? It may help with readability since aliases will be declared closer to where hps are initialized (as opposed to maintaining a list within the `validate` method itself). Similarly, moving logic to `_declare_alias()` will help with keeping code modular.

What do you think?

src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
```diff
@@ -244,15 +244,26 @@ def _visit(name, visited, stack):
...
                 if self.hyperparameters[hp].required:
                     raise exc.UserError("Missing required hyperparameter: {}".format(hp))
                 elif self.hyperparameters[hp].default is not None:
                     user_hyperparameters[hp] = self.hyperparameters[hp].default
 
-        # NOTE: 1. Convert hyperparameters.
+        # NOTE: 2. Convert hyperparameters.
         converted_hyperparameters = {}
         for hp, value in user_hyperparameters.items():
             try:
```
rizwangilani, line 269: With the above change, you could perhaps check for the alias here

test/unit/algorithm_mode/test_algorithm_mode.py
```diff
@@ -116,6 +116,66 @@ def test_hyperparameters6(self):
         hps = hpv.initialize(self.metrics)
         hps.validate(hyperparameters)
 
+    def test_hyperparameters7(self):
```
rizwangilani, line 119: These tests look good. 

However, they are not testing the validate method itself (as the expectation for these tests is that validate works fine). For example, the test wont fail even if we just return an empty dictionary as validated hps. We should be testing the validate method here with test cases for your changes:
https://github.com/aws/sagemaker-xgboost-container/blob/master/test/unit/algorithm_toolkit/test_hyperparameter_validation.py

src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
salmankhurshid1: I'm not sure how we can use `_declare_alias()` here, but I could I add an attribute `alias` to the `Hyperparameter` class. Through this we can declare aliases for our hyperparameters when we are initializing in https://github.com/aws/sagemaker-xgboost-container/blob/master/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py.

src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
salmankhurshid1: Yes, now with the above change, we can check for aliases here along with `# NOTE: 0`.

test/unit/algorithm_mode/test_algorithm_mode.py
salmankhurshid1: This makes sense.
I'll have to add additional unit tests to test the validate method in the file you've linked.

src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
rizwangilani: my bad, I meant `declare_alias(key_name, alias_name)`, which would eventually add an alias to the `alias` attribute from the file you pointed out.

src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
rizwangilani: sounds good.

test/unit/algorithm_mode/test_algorithm_mode.py
rizwangilani: yes, that sounds good.

src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
```diff
@@ -243,16 +244,34 @@ def _visit(name, visited, stack):
 
         return dependencies_stack
 
+    def _declare_alias(self, key_name, alias_name):
+        if key_name not in self.hyperparameters:
+            raise exc.AlgorithmError("Key name { " + key_name + " } does not exist in list of hyperparameters")
```
rizwangilani, line 249: please use the formatter to format the string. See other exceptions in the codebase.

src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
```diff
@@ -162,12 +163,12 @@ def interaction_constraints_validator(value, dependencies):
...
-        hpv.ContinuousHyperparameter(name="lambda", range=hpv.Interval(min_closed=0), required=False,
-                                     tunable=True,
+        hpv.ContinuousHyperparameter(name="lambda", range=hpv.Interval(min_closed=0),
+                                     required=False, tunable=True,
                                      tunable_recommended_range=hpv.Interval(min_closed=0, max_closed=1000,
                                                                             scale=hpv.Interval.LINEAR_SCALE)),
-        hpv.ContinuousHyperparameter(name="alpha", range=hpv.Interval(min_closed=0), required=False,
-                                     tunable=True,
+        hpv.ContinuousHyperparameter(name="alpha", range=hpv.Interval(min_closed=0),
+                                     required=False, tunable=True,
```
rizwangilani, line 171: are these changes still necessary? I think moving `required` to newline is not needed anymore since you are not using the `alias` parameter.

src/sagemaker_algorithm_toolkit/hyperparameter_validation.py
```diff
@@ -243,16 +244,34 @@ def _visit(name, visited, stack):
 
         return dependencies_stack
 
+    def _declare_alias(self, key_name, alias_name):
```
rizwangilani, line 247: by convention, I would not prefix this method with an `_` as this is not a helper method and is invoked outside of the class.

test/unit/algorithm_toolkit/test_hyperparameter_validation.py
```diff
@@ -187,3 +187,72 @@ def test_simple_tuple(self):
...
         self.assertEqual(result["tuple_"], (1, 0, -1))
+
+    def test_tuple_alias(self):
+        hyperparameters = hpv.Hyperparameters(hpv.TupleHyperparameter(name="tuple_",
+                                                                           range=[-1, 0, 1],
+                                                                           required=False))
+        hyperparameters._declare_alias("tuple_", "tup_")
+
+        with self.assertRaises(exc.UserError):
+            hyperparameters.validate({"tup_": "(1,0,-1,2)"})
```
rizwangilani, line 198: is this test truly required for this code change? Why do we need to test the range parameter?

test/unit/algorithm_toolkit/test_hyperparameter_validation.py
```diff
@@ -187,3 +187,72 @@ def test_simple_tuple(self):
...
+        hps = hpv.Hyperparameters(hpv.ContinuousHyperparameter(name="gamma",
+                                                               range=hpv.Interval(min_closed=0, max_closed=1),
+                                                               required=True))
+        hps._declare_alias("gamma", "min_split_loss")
+
+        with self.assertRaises(exc.UserError):
+            hps.validate({})
+
+        with self.assertRaises(exc.UserError):
+            hps.validate({"min_split_loss": "ha"})
```
rizwangilani, line 213: same comment as above - is this test testing functionality introduced as a part of this commit/code change?

PR #111, opened 2020-05-25
feature: add selectable inference content for csv, json, jsonlines, and recordio-protobuf

wiltonwu:
*Description of changes:*
* Adds new selectable inference content feature for accept types: `application/json`, `application/jsonlines`, `application/x-recordio-protobuf`, and `text/csv`
* This new selectable inference content is turned on via the `SAGEMAKER_INFERENCE_OUTPUT` enviornment variable.
  * If that env var is present, the env var will determine the content in the inference response in a string of content keys separated by commas corresponds to the allowed selectable content
  * e.g. `SAGEMAKER_INFERENCE_OUTPUT='predicted_label,probabilities'`

Tested with `tox` and integration tests.


src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -266,6 +303,15 @@ def invocations():
...
         return flask.Response(response="Unable to evaluate payload provided: %s" % e, status=http.client.BAD_REQUEST)
 
+    if serve_utils.is_selectable_inference_response():
+        try:
+            accept = _parse_accept(flask.request)
+        except Exception as e:
+            logging.exception(e)
+            return flask.Response(response=str(e), status=http.client.NOT_ACCEPTABLE)
+
+        return _handle_selectable_inference_response(preds, accept)
```
ericangelokim, line 313: This seems a little different from what was discussed; I thought the selectable response would only be generated based on an env var that is set. However, we are parsing for configs that might not be used regardless of env vars that is set.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -0,0 +1,254 @@
...
+from sagemaker_xgboost_container.algorithm_mode import integration
+
+
+logging = integration.setup_main_logger(__name__)
+
+SAGEMAKER_INFERENCE_OUTPUT = 'SAGEMAKER_INFERENCE_OUTPUT'
+
+# xgboost objective learning tasks
+# https://xgboost.readthedocs.io/en/release_0.90/parameter.html#learning-task-parameters
+REG_SQUAREDERR = 'reg:squarederror'
```
ericangelokim, line 31: Let's move these constants, since these are the same values regardless of algorithm mode or script mode. There is a directory for constants already.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -0,0 +1,254 @@
...
+
+from sagemaker_xgboost_container.algorithm_mode import integration
+
+
+logging = integration.setup_main_logger(__name__)
+
+SAGEMAKER_INFERENCE_OUTPUT = 'SAGEMAKER_INFERENCE_OUTPUT'
+
+# xgboost objective learning tasks
+# https://xgboost.readthedocs.io/en/release_0.90/parameter.html#learning-task-parameters
```
ericangelokim, line 30: 1.0 now, not 0.90?

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -0,0 +1,254 @@
...
+        return [0, 1]
+    if "multi:" in objective and num_class:
+        return list(range(int(num_class)))
+    return np.nan
+
+
+def _get_predicted_label(objective, data):
+    if objective in [BINARY_HINGE, MULTI_SOFTMAX]:
+        return data
+    if objective in [BINARY_LOG]:
```
ericangelokim, line 96: This makes me nervous, since we are now changing the value returned to customers. I don't have intuition on if this is the right behavior to implement. Please explain reasoning.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -0,0 +1,254 @@
...
+        for key in item.keys():
+            value = item[key] if type(item[key]) is list else [item[key]]
+            _write_record(record, key, value)
+        record_bio.write(record.SerializeToString())
+        record.Clear()
+        _write_recordio(recordio_bio, record_bio.getvalue())
+    return recordio_bio.getvalue()
+
+
+def encode_selected_content(content, keys, accept):
```
ericangelokim, line 218: We also have encoder code that should be useable by customers during script mode, please move this to that submodule.

src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -242,6 +251,34 @@ def _parse_content_data(request):
     return dtest, content_type
 
 
+def _parse_accept(request):
+    accept = request.headers.get("accept").lower().split(";")[0]
+    if not any(accept in mimetypes for mimetypes in ["application/json", "application/jsonlines",
```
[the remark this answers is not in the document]
wiltonwu: Algorithm mode currently just supports CSV, but this change would add support for CSV, JSON, JSONLINES, and RECORDIO-PROTOBUF in algorithm mode with selectable inference turned on. The accept types come from common formats for our other algorithms (i.e. linear learner)

src/sagemaker_xgboost_container/algorithm_mode/serve.py
wiltonwu: This line is only hit if `serve_utils.is_selectable_inference_response()` is `True`, which does the env var check. I may be missing something but what extraneous configs are you referring to when that env var is not set?

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
wiltonwu: This behavior is the same as the logic in the [sklearn xgboost API](https://xgboost.readthedocs.io/en/release_0.90/python/python_api.html#module-xgboost.sklearn). The `XGBClassifier` has similar logic in its [`.predict(..)`](https://github.com/dmlc/xgboost/blob/release_1.0.0/python-package/xgboost/sklearn.py#L892-L896) which works for all classification objective functions. The code for determining predicted labels is based on that sklearn xgboost wrapper.

wiltonwu:
> It is alarming that we are now dictating what the correct values to return to customers are, for example get label for binary_log. The decisions you made make sense to me, but we might want to involve product to make sure this behavior is ok to push out.

Okay! Let's sync up on the product point. Most of the "correct value" decisions are just an extension of the [xgboost sklearn API](https://github.com/dmlc/xgboost/blob/release_1.0.0/python-package/xgboost/sklearn.py)

> Also, were integration tests run with the built image?

Yep! Correct

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
wiltonwu: These encoders are a little niche because they're supposed to be used specifically with the layout of the `content` argument. I think it makes more sense to keep them in the `serve_utils` submodule

src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -21,11 +22,11 @@
 import flask
 import gunicorn.app.base
 
-from sagemaker_xgboost_container.algorithm_mode import integration, serve_utils
+from sagemaker_xgboost_container.algorithm_mode import integration
+from sagemaker_xgboost_container.algorithm_mode import serve_utils
 from sagemaker_xgboost_container.constants import sm_env_constants
 
 
-SAGEMAKER_BATCH = os.getenv("SAGEMAKER_BATCH")
```
ericangelokim, line 28: Why did you remove this? This was previously parsed once during env initialization, but now this gets parsed every inference call. Add back.

src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -145,6 +154,51 @@ def execution_parameters():
...
+
+def _handle_selectable_inference_response(predictions, accept):
+    """Retrieve the additional prediction data for selectable inference mode.
+
+    :param predictions: output of xgboost predict (list of numpy objects)
+    :param accept: requested accept type (str)
+    :return: flask response with encoded predictions
+    """
+    try:
+        # get objective function and content keys needed to make output
```
ericangelokim, line 183: These comments, while well intentioned, are unnecessary. The clearest code should not need inline documentation. Remove.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -110,3 +163,269 @@ def predict(booster, model_format, dtest, input_content_type):
...
+    return SAGEMAKER_INFERENCE_OUTPUT in os.environ
+
+
+def get_selected_output_keys():
+    """Get the selected output content keys from the `SAGEMAKER_INFERENCE_OUTPUT` env var.
+
+    :return: selected output content keys (list of str)
+    """
+    if is_selectable_inference_output():
+        return os.getenv(SAGEMAKER_INFERENCE_OUTPUT).replace(' ', '').lower().split(',')
```
ericangelokim, line 178: Parse this once at top, rather than call it every invocation.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -110,3 +163,269 @@ def predict(booster, model_format, dtest, input_content_type):
...
+        return max(raw_prediction).item()
+    if objective in [BINARY_LOG]:
+        return raw_prediction.item()
+    return np.nan
+
+
+def _get_probabilities(objective, raw_prediction):
+    """Get the probability scores for all classes for a classification problem.
+
+    A probability score is a value between 0.0 and 1.0.
```
ericangelokim, line 237: Unnecessary. The important part was that this gets scores for all classes.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -110,3 +163,269 @@ def predict(booster, model_format, dtest, input_content_type):
...
+        return int(raw_prediction > 0)
+    if objective in [MULTI_SOFTPROB]:
+        return np.argmax(raw_prediction).item()
+    return np.nan
+
+
+def _get_probability(objective, raw_prediction):
+    """Get the probability score for a classification problem.
+
+    A probability score is a value between 0.0 and 1.0. In binary classification,
```
ericangelokim, line 218: Just leave the critical info here, which is that this is the prob of the winning class.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -110,3 +163,269 @@ def predict(booster, model_format, dtest, input_content_type):
...
+        classone_probs = raw_prediction.item()
+        classzero_probs = 1.0 - classone_probs
+        return [classzero_probs, classone_probs]
+    return np.nan
+
+
+def _get_raw_score(objective, raw_prediction):
+    """Get the raw score for a classification problem.
+
+    A raw score is defined as any numeric value. The higher the value, the more likely
```
ericangelokim, line 255: Same comments as above.

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -110,3 +163,269 @@ def predict(booster, model_format, dtest, input_content_type):
...
+            output[PREDICTED_SCORE] = raw_prediction.item()
+        if invalid_selected_keys:
+            for invalid_selected_key in invalid_selected_keys:
+                output[invalid_selected_key] = np.nan
+        predictions.append(output)
+    return predictions
+
+
+def _encode_selected_predictions_csv(predictions, ordered_keys_list):
+    """Encode predictions is csv format.
```
ericangelokim, line 353: Nit: in, not is

src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py
```diff
@@ -110,3 +163,269 @@ def predict(booster, model_format, dtest, input_content_type):
...
+            value = item[key] if type(item[key]) is list else [item[key]]
+            _write_record(record, key, value)
+        record_bio.write(record.SerializeToString())
+        record.Clear()
+        _write_recordio(recordio_bio, record_bio.getvalue())
+    return recordio_bio.getvalue()
+
+
+def encode_selected_predictions(predictions, selected_content_keys, accept):
+    """Encodes the selected predictions and keys based on the given accept type.
```
ericangelokim, line 409: Nit: Encode

test/unit/algorithm_mode/test_serve_utils.py
```diff
@@ -76,3 +82,180 @@ def test_predict_valid_content_type(correct_content_type):
...
+
+
+def test_is_selectable_inference_response_true():
+    os.environ['SAGEMAKER_INFERENCE_OUTPUT'] = 'predicted_label'
+    assert serve_utils.is_selectable_inference_output()
+    del os.environ['SAGEMAKER_INFERENCE_OUTPUT']
+
+
+def test_get_selected_content_keys():
+    os.environ['SAGEMAKER_INFERENCE_OUTPUT'] = 'predicted_label'
```
ericangelokim, line 118: Instead of changing this value in the environment, you should be using a monkey patch or similar for this test.

test/unit/algorithm_mode/test_serve_utils.py
```diff
@@ -76,3 +82,180 @@ def test_predict_valid_content_type(correct_content_type):
...
+@pytest.mark.parametrize('objective, predictions, expected_raw_score', [
+    (serve_utils.BINARY_LOG, np.float64(0.6), 0.6),
+    (serve_utils.MULTI_SOFTPROB, np.array([0.1, 0.5, 0.4]), 0.5),
+    (serve_utils.BINARY_LOGRAW, np.float64(-7.6), -7.6)
+])
+def test_get_raw_score(objective, predictions, expected_raw_score):
+    assert serve_utils._get_raw_score(objective, predictions) == expected_raw_score
+
+
+def est_get_raw_score_nan():
```
ericangelokim, line 209: ```test_get_raw_score_nan()```

src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -145,6 +155,48 @@ def execution_parameters():
     return flask.Response(response=response_text, status=http.client.OK, mimetype="application/json")
 
 
+def _parse_accept(request):
+    """Gets the accept type for a given request.
```
ericangelokim, line 159: Nit: Get

src/sagemaker_xgboost_container/encoder.py
```diff
@@ -107,6 +109,24 @@ def recordio_protobuf_to_dmatrix(string_like):  # type: (bytes) -> xgb.DMatrix
     xgb_content_types.X_RECORDIO_PROTOBUF: recordio_protobuf_to_dmatrix}
 
 
+def json_to_jsonlines(json_data):
+    """Utility function to convert a json response to a jsonlines response.
```
ericangelokim, line 113: Nit: Convert a json response to jsonlines

PR #122, opened 2020-06-08
fix: change parse accept type logic when accept not set

wiltonwu:
*Description of changes:*
* This change is to fix the behavior when `accept` is not specified in the header of the request (occurs during batch transform)
  * If `accept` is not specified or "\*/\*", it should default to `SAGEMAKER_DEFAULT_INVOCATIONS_ACCEPT` env var
* Previous logic would cause a `'NoneType' object has no attribute 'lower'` error

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/serve.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/serve.py
@@ -28,7 +28,7 @@ from sagemaker_xgboost_container.constants import sm_env_constants
 
 
 SAGEMAKER_BATCH = os.getenv(sm_env_constants.SAGEMAKER_BATCH)
-SUPPORTED_ACCEPT = ["application/json", "application/jsonlines", "application/x-recordio-protobuf", "text/csv"]
+SUPPORTED_ACCEPTS = ["application/json", "application/jsonlines", "application/x-recordio-protobuf", "text/csv"]
 logging = integration.setup_main_logger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -164,12 +165,11 @@ def _parse_accept(request):
...
     :return: parsed accept type
     """
-    accept, _ = cgi.parse_header(request.headers.get("accept").lower())
-    if not any(accept in mimetypes for mimetypes in ["application/json", "application/jsonlines",
-                                                     "application/x-recordio-protobuf", "text/csv"]):
-        if sm_env_constants.SAGEMAKER_DEFAULT_INVOCATIONS_ACCEPT in os.environ:
-            return os.getenv('SAGEMAKER_DEFAULT_INVOCATIONS_ACCEPT')
-        raise ValueError("Accept type {} is not supported.".format(accept))
+    default_accept = os.getenv('SAGEMAKER_DEFAULT_INVOCATIONS_ACCEPT')
+    accept, _ = cgi.parse_header(request.headers.get("accept", default_accept))
```
aws-patlin, line 169: Does the result of `request.headers.get("accept", default_accept)` need to be lowercased still?

wiltonwu: Thanks Patrick, updated to lowercase and to improve the error handling.

PR #123, opened 2020-06-11
fix: bug for custom metric calculations for multi:softprob

wiltonwu:
*Description of changes:*
* bug fix for custom accuracy metric calculations for `multi:softprob`
* While running benchmarks on HPO, we ran into an issue where multiclass problem types would perform significantly worse with `multi:softprob` compared to using the `multi:softmax` objective
* also affects `binary:logistic`

### DIFF
--- a/test/unit/algorithm_mode/test_custom_metrics.py
+++ b/test/unit/algorithm_mode/test_custom_metrics.py
@@ -44,12 +44,19 @@ multiclass_train_data = np.random.rand(10, 2)
 multiclass_train_label = np.array([0, 0, 1, 1, 1, 1, 1, 2, 2, 2])
 multiclass_dtrain = xgb.DMatrix(multiclass_train_data, label=multiclass_train_label)
 multiclass_preds = np.ones(10)
+multiclass_preds_softprob = np.asarray([[0.8, 0.1, 0.1]] * 10)
 
 
 def test_multiclass_accuracy():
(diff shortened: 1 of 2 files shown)


test/unit/algorithm_mode/test_custom_metrics.py
```diff
@@ -52,7 +53,19 @@ def test_multiclass_accuracy():
     assert accuracy_result == .5
 
 
+def test_multiclass_accuracy_softprob():
```
pdasamzn, line 56: I would add an explicit test for binary classification.

wiltonwu: Added

test/unit/algorithm_mode/test_custom_metrics.py
```diff
@@ -44,6 +44,7 @@ def test_mse():
 multiclass_train_label = np.array([0, 0, 1, 1, 1, 1, 1, 2, 2, 2])
 multiclass_dtrain = xgb.DMatrix(multiclass_train_data, label=multiclass_train_label)
 multiclass_preds = np.ones(10)
+multiclass_preds_softprob = np.asarray([[0.8, 0.1, 0.1]] * 10)
```
pdasamzn, line 60: Please take the array with different values in each row, to ensure that behaving as expected.

wiltonwu: Sure, updated the values.

PR #126, opened 2020-07-07
Add transform_fn

edwardjkim:
*Description of changes:*

We've had customer feedback that `transform_fn` (where customers can combine `input_fn`, `predict_fn` and `output_fn` in a single function to handle inference requests) is lacking in xgboost. This PR enables the use of `transform_fn` in script mode.

The actual code change to enable `transform_fn` is minimal in `_user_module_transformer()` of [src/sagemaker_xgboost_container/serving.py](https://github.com/aws/sagemaker-xgboost-container/pull/126/files#diff-52485661f881d7500a690e18e9a27dc4), so the review can be focused on `serving.py`. The rest of the changes are for adding container tests for [xgboost abalone](https://github.com/awslabs/amazon-sagemaker-examples/tree/master/introduction_to_amazon_algorithms/xgboost_abalone) script. The sample script at `test/resources/abalone/abalone_distributed.py` is the same script from [sagemaker examples](https://github.com/awslabs/amazon-sagemaker-examples/tree/master/introduction_to_amazon_algorithms/xgboost_abalone) for xgboost distributed training.

After this PR is merged, the following text will be added to the [Python SDK documentation](https://sagemaker.readthedocs.io/en/stable/frameworks/xgboost/using_xgboost.html#sagemaker-xgboost-model-server) after the section on `input_fn`, `predict_fn`, and `output_fn`:

> If you would rather not structure your code around the three methods described above, you can instead define your own `transform_fn` to handle inference requests. An error is thrown if a `transform_fn` is present in conjunction with any `input_fn`, `predict_fn`, and/or `output_fn`. `transform_fn` has the following signature:
> 
> ```def transform_fn(model, request_body, content_type, accept_type)```
>
> where model is the `model` objecte loaded by `model_fn`, `request_body` is the data from the inference request, `content_type` is the content type of the request, and `accept_type` is the request content type for the response.
>
> This one function should handle processing the input, performing a prediction, and processing the output. The return object should have the response data and accept_type (the content type of the response data).

*Testing*
New container tests in this PR.
Existing integration tests succeeded.


src/sagemaker_xgboost_container/serving.py
```diff
@@ -88,13 +97,26 @@ def default_output_fn(prediction, accept):
...
-
-    return transformer.Transformer(model_fn=model_fn, input_fn=input_fn, predict_fn=predict_fn,
-                                   output_fn=output_fn)
+    model_fn = getattr(user_module, "model_fn", default_model_fn)
+    input_fn = getattr(user_module, "input_fn", None)
+    predict_fn = getattr(user_module, "predict_fn", None)
+    output_fn = getattr(user_module, "output_fn", None)
+    transform_fn = getattr(user_module, "transform_fn", None)
+
+    if transform_fn and (input_fn or predict_fn or output_fn):
```
balajitummala, line 106: Transformer class is already throwing a [ValueError](https://github.com/aws/sagemaker-containers/blob/master/src/sagemaker_containers/_transformer.py#L154) for the same condition so curious if this would be redundant.

edwardjkim: The logic in this PR is structured so that the Transformer won't enter that if statement. Also, I thought it should be raised as a customer error on our side rather than a generic ValueError.

PR #129, opened 2020-08-06
Add Non-numeric data check

balajitummala:
*Description of changes:*
Currently, XGBoost doesn't fail when reading non-numeric data in CSV. Added a check to throw an error if the training data contains non-numeric data.

**Testing**
- Added a unit test
tox -e py3-xgboost1.0,flake8


edwardjkim:
Doesn't upstream XGBoost allow non-numeric data, for example NA for missing values?

PR #133, opened 2020-09-04
Release xgboost 1.2 with GPU support

edwardjkim:
*Description of changes:*

This CR upgrades XGBoost to [1.2](https://github.com/dmlc/xgboost/releases/tag/v1.2.0) and enables GPU support.
- When the image is build with xgboost 1.1, many integration tests fail with an xgboost error on feature mismatch, e.g.,
  ```
  xgboost.core.XGBoostError: [16:47:33] /workspace/src/learner.cc:1062: Check failed: learner_model_param_.num_feature == p_fmat->Info().num_col_ (9 vs. 8) : Number of columns does not match number of features in booster.
  ```
  This is due to a bug in 1.1 (github issues: https://github.com/dmlc/xgboost/issues/5841, https://github.com/dmlc/xgboost/issues/5709). This has been fixed in 1.2 (https://github.com/dmlc/xgboost/pull/5955). From [1.2 release notes](https://github.com/dmlc/xgboost/releases/tag/v1.2.0):
  ```
  Restore capability to run prediction when the test input has fewer features than the training data (#5955). This capability is necessary to support predicting with LIBSVM inputs. The previous release (1.1) had broken this capability, so we restore it in this version with better tests.
  ```
  Since it doesn't sound like upstream XGBoost will backport this fix to 1.1, we release 1.2 in this CR.
- New in XGBoost 1.1 & 1.2
  - `silent` parameter has been removed in 1.1 in favor of `verbose`.
  - A new objective `survival:aft` is added to support survival analysis: https://xgboost.readthedocs.io/en/latest/tutorials/aft_survival_analysis.html
- MLIO needs to be upgraded. The latest version of MLIO is v0.6. However, the conda package for v0.5 and v0.6 add ~3GB uncompressed (~1GB compressed) to the docker image (mainly due to a huge list of dependencies for image reader, e.g., ffmpeng, opencv, which were newly added in v0.5) increasing training time by ~1 minute. Thus, Dockerfile is optimized and rewritten to install mlio from source. The final image size is 1326.14 MB (compressed) with XGBoost 1.2, MLIO upgrade, and GPU support, compared to 1225.65 MB (compressed) for `1.0-1-cpu-py3`.
- GPU support
  - We could install the CUDA toolkit, but installing CUDA Toolkit will increase the image size by around 700 MB (compressed). The proposed base image `nvidia/cuda:${CUDA_VERSION}-base-ubuntu${UBUNTU_VERSION}` is a small image that contains a minimal set of CUDA runtime files.
  - Customers will have to specify the parameter `tree_method: gpu_hist` (and use the correct instance type, e.g., `p3.xlarge`, `p3.2xlarge`) to enable GPU training.
- With GPU support in the same image as the CPU image, it is no longer necessary to append the architecture in the image tag. Since we dropped Python 2 support, the `-cpu-py3` in the framework version is also redundant, and this CR proposes to drop the `-<architecture>-<python version>` suffix. (However, we will keep the old tag format in the deployment pipelines for backwards compatibility. That is, we will tag the same image with two tags: `1.2-1` and `1.2-1-cpu-py3`.)

Testing: tox, integration tests


edwardjkim:
Codebuild failed, but I'm sending this PR out for review while I work on fixing codebuild.

edwardjkim:
Closing this PR since we had to reset codebuild. I'll make a new PR.

PR #140, opened 2020-10-26
Pin conda installation file for 1.2-1

edwardjkim:
*Description of changes:*

Pins Python version in Conda. Without pinning the Python version, a rebuild of the image will bump the Python version when conda-latest bumps the Python version.

Also pins versions for all major dependencies. We have had too many issues related to Conda/Python/libraries getting upgraded inadvertently when a CodeBuild build is triggered. This PR pins the versions in `requirements.txt` to a specific version or puts an upper bound.

Duplicate entries in `test-requirements.txt` are removed because it is unnecessary have the libraries listed in  `requirements.txt` again in `test-requirements.txt`.

The versions were inspected with `pip freeze`:
```
docker run 246618743249.dkr.ecr.us-west-2.amazonaws.com/sagemaker-xgboost:0.90-2-cpu-py3 sh -c 'python -m pip freeze'
```

See also https://github.com/aws/sagemaker-xgboost-container/pull/139.

### DIFF
--- a/docker/1.2-1/base/Dockerfile.cpu
+++ b/docker/1.2-1/base/Dockerfile.cpu
@@ -3,6 +3,9 @@ ARG CUDA_VERSION=10.2
 
 FROM nvidia/cuda:${CUDA_VERSION}-base-ubuntu${UBUNTU_VERSION}
 
+ARG CONDA_VERSION=4.8.3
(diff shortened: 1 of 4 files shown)


edmrossi (approved):
🚢

PR #143, opened 2020-10-26
Pin dependency versions for 1.0-1

edwardjkim:
*Description of changes:*

Pins versions for all major dependencies. We have had too many issues related to Conda/Python/libraries getting upgraded inadvertently when a CodeBuild build is triggered. This PR pins the versions in `requirements.txt` to a specific version or puts an upper bound.

The versions were inspected with `pip freeze`:
```
docker run 246618743249.dkr.ecr.us-west-2.amazonaws.com/sagemaker-xgboost:1.0-1-cpu-py3 sh -c 'python -m pip freeze'
```

See also https://github.com/aws/sagemaker-xgboost-container/pull/140.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,16 +1,19 @@
+Flask==1.1.1  # sagemaker-containers requires flask 1.1.1
 PyYAML<4.3
+boto3==1.10.14
+botocore==1.13.14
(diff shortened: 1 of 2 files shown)


edmrossi (approved):
🚢

PR #157, opened 2021-02-03
Uninstall typing for smdebug version bump

edwardjkim:
*Description of changes:*

`sagemaker-containers` (which is deprecated) installs `typing` which was necessary in Python 3.5 for typing support but is no longer necessary in Python 3.7. Install `typing` in Python 3.7 can result in conflicts, e.g., when installing `python3 -m pip install black`. The simplest solution is to uninstall `typing` at the end.


edwardjkim:
We are hitting docker pull rate limit:
```
Step 4/23 : FROM nvidia/cuda:${CUDA_VERSION}-base-ubuntu${UBUNTU_VERSION}@sha256:${IMAGE_DIGEST}
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

aws-patlin (approved):

PR #158, opened 2021-02-13
cross validation for autopilot

lihaife:
*Issue #, if available:*

*Description of changes:*
cross validation for autopilot

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
@@ -123,7 +123,7 @@ def initialize(metrics):
         hpv.IntegerHyperparameter(name="early_stopping_rounds", range=hpv.Interval(min_closed=1), required=False),
         hpv.CategoricalHyperparameter(name="booster", range=["gbtree", "gblinear", "dart"], required=False),
         hpv.IntegerHyperparameter(name="verbosity", range=hpv.Interval(min_closed=0, max_closed=3), required=False),
-        hpv.IntegerHyperparameter(name="nfold", range=hpv.Interval(min_closed=2), required=False, tunable=False),
+        hpv.IntegerHyperparameter(name="_nfold", range=hpv.Interval(min_closed=2), required=False, tunable=False),
         hpv.IntegerHyperparameter(name="nthread", range=hpv.Interval(min_closed=1), required=False),
         hpv.ContinuousHyperparameter(name="eta", range=hpv.Interval(min_closed=0, max_closed=1), required=False,
                                      tunable=True,
--- a/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
@@ -123,6 +123,7 @@ def initialize(metrics):
         hpv.IntegerHyperparameter(name="early_stopping_rounds", range=hpv.Interval(min_closed=1), required=False),
         hpv.CategoricalHyperparameter(name="booster", range=["gbtree", "gblinear", "dart"], required=False),
         hpv.IntegerHyperparameter(name="verbosity", range=hpv.Interval(min_closed=0, max_closed=3), required=False),
+        hpv.IntegerHyperparameter(name="nfold", range=hpv.Interval(min_closed=2), required=False, tunable=False),
         hpv.IntegerHyperparameter(name="nthread", range=hpv.Interval(min_closed=1), required=False),
         hpv.ContinuousHyperparameter(name="eta", range=hpv.Interval(min_closed=0, max_closed=1), required=False,
                                      tunable=True,
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -221,9 +221,19 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir, 
         logging.info("Validation matrix has {} rows".format(val_dmatrix.num_row()))
 
     try:
+        nfold = train_cfg.pop("nfold", None)
+
         bst = xgb.train(train_cfg, train_dmatrix, num_boost_round=num_round, evals=watchlist, feval=configured_feval,
                         early_stopping_rounds=early_stopping_rounds, callbacks=callbacks, xgb_model=xgb_model,
                         verbose_eval=False)
+
+        if nfold is not None:
+            all_dmatrix = train_dmatrix
+            # TODO: XGBoost doesn't support DMatrix concatenation.
+            #if val_dmatrix is not None:
+            #    all_dmatrix = bind(train_dmatrix, val_dmatrix)
+            bst = xgb.cv(train_cfg, all_dmatrix, nfold=nfold, num_boost_round=num_round,
+                         feval=configured_feval, show_stdv=False, verbose_eval=True)
     except Exception as e:
         for customer_error_message in CUSTOMER_ERRORS:
             if customer_error_message in str(e):
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -232,8 +232,8 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir, 
             # TODO: XGBoost doesn't support DMatrix concatenation.
             #if val_dmatrix is not None:
             #    all_dmatrix = bind(train_dmatrix, val_dmatrix)
-            bst = xgb.cv(train_cfg, all_dmatrix, nfold=nfold, num_boost_round=num_round,
-                         feval=configured_feval, show_stdv=False, verbose_eval=True)
+            xgb.cv(train_cfg, all_dmatrix, nfold=nfold, num_boost_round=num_round,
+                   feval=configured_feval, show_stdv=False, verbose_eval=True)
     except Exception as e:
         for customer_error_message in CUSTOMER_ERRORS:
             if customer_error_message in str(e):
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -233,7 +233,8 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir, 
             # if val_dmatrix is not None:
             #     all_dmatrix = bind(train_dmatrix, val_dmatrix)
             xgb.cv(train_cfg, all_dmatrix, nfold=nfold, num_boost_round=num_round,
-                   feval=configured_feval, show_stdv=False, verbose_eval=True)
+                   feval=configured_feval, early_stopping_rounds=early_stopping_rounds, show_stdv=False,
+                   verbose_eval=True)
     except Exception as e:
         for customer_error_message in CUSTOMER_ERRORS:
             if customer_error_message in str(e):
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -261,4 +261,3 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
         with open(model_location, 'wb') as f:
             pkl.dump(bst, f, protocol=4)
         logging.debug("Stored trained model at {}".format(model_location))
-
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -86,7 +86,7 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
     val_dmatrix = get_dmatrix(validate_path, content_type, is_pipe=is_pipe) \
         if val_files_size > 0 else None
 
-    train_val_dmatrix = train_dmatrix
+    train_val_dmatrix = None
     if combine_train_val and train_dmatrix is not None and val_dmatrix is not None:
         logging.info("Read both train and validation data into one DMatrix")
         train_val_dmatrix = get_dmatrix([train_path, validate_path], content_type,
@@ -132,7 +132,7 @@ def sagemaker_train(train_config, data_config, train_path, val_path, model_dir, 
     is_pipe = (input_mode == Channel.PIPE_MODE)
 
     validation_channel = validated_data_config.get('validation', None)
-    combine_train_val = 'nfold' in validated_train_config
+    combine_train_val = '_nfold' in validated_train_config
     train_dmatrix, val_dmatrix, train_val_dmatrix = get_validated_dmatrices(train_path, val_path, file_type,
                                                                             csv_weights, is_pipe, combine_train_val)
 
@@ -233,7 +233,7 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
         logging.info("Validation matrix has {} rows".format(val_dmatrix.num_row()))
 
     try:
-        nfold = train_cfg.pop("nfold", None)
+        nfold = train_cfg.pop("_nfold", None)
 
         bst = xgb.train(train_cfg, train_dmatrix, num_boost_round=num_round, evals=watchlist, feval=configured_feval,
                         early_stopping_rounds=early_stopping_rounds, callbacks=callbacks, xgb_model=xgb_model,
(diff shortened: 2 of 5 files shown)


src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -221,9 +221,19 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir,
...
+        nfold = train_cfg.pop("nfold", None)
+
         bst = xgb.train(train_cfg, train_dmatrix, num_boost_round=num_round, evals=watchlist, feval=configured_feval,
                         early_stopping_rounds=early_stopping_rounds, callbacks=callbacks, xgb_model=xgb_model,
                         verbose_eval=False)
+
+        if nfold is not None:
+            all_dmatrix = train_dmatrix
+            # TODO: XGBoost doesn't support DMatrix concatenation.
+            #if val_dmatrix is not None:
```
edwardjkim, line 233: CodeBuild is failing due to this line:
```
src/sagemaker_xgboost_container/algorithm_mode/train.py:233:13: E265 block comment should start with '# '
975 | ERROR: InvocationError for command /app/.tox/flake8/bin/flake8 setup.py src test (exited with code 1)
976 | ___________________________________ summary ____________________________________
977 | py37-xgboost1.2: commands succeeded
978 | ERROR:   flake8: commands failed
979
```

edwardjkim (commented):
- If the customer uses the `nfold` hyperparameter, it looks like both `xgb.train()` and `xgb.cv()` will run. Is this intended? I'm having difficulty seeing why we would want to run both. Isn't `cv()` just `train()` with cross validation?
- What happens to checkpointing if nfold is enabled? What happens to early stopping?
- What happens if the customer uses a validation set but uses `nfold`? Is the validation simply ignored? If so, should we warn the the customer in some way?
- Can we remove lines 231-234?

src/sagemaker_xgboost_container/algorithm_mode/train.py
lihaife: Thanks. I fixed it. More importantly, are you aware of anyway to concatenate DMatrix?

lihaife:
@edwardjkim Overall, this feature is designed for Autopilot/HPO, not for general purpose. We won't reveal the `nfold` parameter to public. 

- If the customer uses the nfold hyperparameter, it looks like both xgb.train() and xgb.cv() will run. Is this intended? I'm having difficulty seeing why we would want to run both. Isn't cv() just train() with cross validation?

It is intended for autopilot. First of all, each round of `cv()` uses less data. Besides, it is not clear which fold should be serialized. The "best" one may cause overfitting. Besides, autopilot will allow the user to provide custom train/validation split. If so, the serialized model should use the provided train set.

- What happens to checkpointing if nfold is enabled? What happens to early stopping?

No checkpoint by design as we only care the metric from cross validation. Although `cv()` support early stopping, there is run time warning "Parameters: { early_stopping_rounds } might not be used." Interestingly, the log shows the parameter is in effect. For safety, we removed it.  I enable it in latest commit but we should be careful here.

- What happens if the customer uses a validation set but uses nfold? Is the validation simply ignored? If so, should we warn the the customer in some way?

Ideally, we would like to combine train and validation set for cross validation. However, it seems no way to combine two DMatrix. Please let's know if there is a trick.

- Can we remove lines 231-234?

We will remove it if we can combine train and validation set. Otherwise, I would like to keep it as a TODO reminder.

- Are there functional and/or integration tests?

There is unit test for new hyper-parameter. All integration tests seem for script mode. If we have tests for algorithm mode, I can add cross validation test there. Currently I do the integration test by a local input directory that mimics EASE /opt/ml. BTW, I shared the tgz of setting in slack.

```
docker run -it --user root -p 8080:8080 -v "$(pwd)"/../test/training/abalone:/opt/ml lihaife-xgboost:1.2-1-cpu-py3 train
```

edwardjkim:
Thanks for the context! That helps with understanding the code. Some follow-up comments:
- If `nfold` is intended only for autopilot, I think it should be named so that the intent is super clear. Something like `_nfold_for_autopilot_use_only`. If that seems like overkill, I think it should at least start with an underscore, i.e., `_nfold`, maybe with an inline comment that makes the intended use clear.
- I don't know if there is a way to concatenate DMatrices either. Data are read as files before being converted to DMatrices, so preprocessing the files before converting them to a DMatrix is an option, but it would need a significant refactor.
- If this feature is only intended for autopilot, I think the tests can be on the autopilot side. My main concern was how this is going to work with HPO because if `nfold` is specified, both `train()` and `cv()` will emit metrics and I'm not sure if it's going to mess with HPO. HPO uses regexes to extract the metrics from the logs and the regexes have been fragile in the past.

lihaife:
I cannot see the CodeBuild results. Can you please grant me access?

edwardjkim:
CodeBuild:
```
ERROR:   flake8: commands failed
```
```
src/sagemaker_xgboost_container/algorithm_mode/train.py:264:1: W391 blank line at end of file
```

lihaife:
Thanks @edwardjkim  I got flake8 installed and hopefully tox will show these code style issues.

Besides, I revise data_utils.py a little bit so that we can read train and validate channels into on DMatrix. I add unit test for it and also tested it in local integration test and autopilot end-2-end tests. It works fine for both file mode and pipe mode.

edwardjkim:
Since the latest revision touches upon some critical code paths, can we also run XGBoost integration tests? Integration tests are run internally via the `SMFrameworksXGBoostTests` Brazil package.  It's derived from `AIAlgorithmsQA` and has the interface as ``AIAlgorithmsQA``. Please let me know via internal channels if there are any issues with running ``SMFrameworksXGBoostTests``.

lihaife:
All integration tests pass

```
(21-02-22 19:22:22) <0> [~/workplace/XGBoost/src/SMFrameworksXGBoostTests]  
dev-dsk-lihaife-1e-5aa6fcfe % bb integration --stage=gamma --image=900597767885.dkr.ecr.us-west-2.amazonaws.com/sagemaker-xgboost:lihaife-1.2-1-cpu-py3
---------------------------- generated xml file: /workplace/lihaife/XGBoost/src/SMFrameworksXGBoostTests/build/brazil-integ-tests/TEST-results.xml ----------------------------
- generated html file: /home/lihaife/workplace/XGBoost/build/SMFrameworksXGBoostTests/SMFrameworksXGBoostTests-1.0/AL2_x86_64/DEV.STD.PTHREAD/build/brazil-unit-tests/python3.6.html -
- generated xml file: /home/lihaife/workplace/XGBoost/build/SMFrameworksXGBoostTests/SMFrameworksXGBoostTests-1.0/AL2_x86_64/DEV.STD.PTHREAD/build/brazil-unit-tests/TEST-python3.6.xml -
- generated html file: /home/lihaife/workplace/XGBoost/build/SMFrameworksXGBoostTests/SMFrameworksXGBoostTests-1.0/AL2_x86_64/DEV.STD.PTHREAD/build/brazil-unit-tests/python3.6.html -
======================================================================== 122 passed in 786.80 seconds =========================================================================
BUILD SUCCEEDED
```

lihaife:
`xgb.train()` print metrics in format such as `train-rmse=xxx, validation-rmse=yyy`. But `xgb.cv()` print metrics in format such as `train-rmse=xxx, test-rmse=yyy`. Is there anyway that we can make `cv()` print in the same way as `train()`?

edwardjkim:
> `xgb.train()` print metrics in format such as `train-rmse=xxx, validation-rmse=yyy`. But `xgb.cv()` print metrics in format such as `train-rmse=xxx, test-rmse=yyy`. Is there anyway that we can make `cv()` print in the same way as `train()`?

It's possible to control the output with a custom callback function. The steps would be to write a print function that is similar to [print_evalution()](https://github.com/dmlc/xgboost/blob/ca3da55de47bfe693b50dd9000c4c84a1cc90180/python-package/xgboost/callback.py#L37) and pass it as a callback by invoking `cv(..., callbacks=[print_evaluation], verbaose_eval=False)` and suppressing the default output with `verbose_eval=False`.

lihaife:
Thanks @edwardjkim. We eventually decide to keep them different so that regular validation metric and cross validation metric report won't be mixed.

src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
```diff
@@ -123,6 +123,7 @@ def interaction_constraints_validator(value, dependencies):
         hpv.IntegerHyperparameter(name="early_stopping_rounds", range=hpv.Interval(min_closed=1), required=False),
         hpv.CategoricalHyperparameter(name="booster", range=["gbtree", "gblinear", "dart"], required=False),
         hpv.IntegerHyperparameter(name="verbosity", range=hpv.Interval(min_closed=0, max_closed=3), required=False),
+        hpv.IntegerHyperparameter(name="nfold", range=hpv.Interval(min_closed=2), required=False, tunable=False),
```
edwardjkim, line 126: If `nfold` is internal for autopilot use, please mark it for internal use by renaming it to `_nfold`. See `_tuning_objective_metric`: https://github.com/aws/sagemaker-xgboost-container/blob/master/src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py#L216.

With our lack of thorough documentation, customers often resort to code, so they will start using this. Most of them will just try it and see that it doesn't work as expected and won't report the issue. Those that do will be filtered incorrectly by SAs who again refer to code to give (incorrect) advice. If it finally reaches our team, our only explanation will be that it's meant for internal use.

I guess it's ultimately Rahul. Product, and/or your call with naming, but I feel pretty strongly about it.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -299,7 +299,8 @@ def _get_csv_dmatrix_pipe_mode(pipe_path, csv_weights):
     :return: xgb.DMatrix or None
     """
     try:
-        dataset = [mlio.SageMakerPipe(pipe_path)]
```
edwardjkim, line 302: The existing code (and the refactored code) staring here is not easy to follow with all the nested if statements. Were there any tests for this function? I can't seem to find any tests that test this function.

src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
lihaife: Although it is for autopilot for now, we may enable it for public usage in the future when we have requests from customers.

src/sagemaker_xgboost_container/data_utils.py
lihaife: We have test in integration tests.

src/sagemaker_xgboost_container/algorithm_mode/hyperparameter_validation.py
edwardjkim: Then we can change it back to `nfold` when we enable it for public usage. It should be marked for internal use if the current target is only autopilot.

src/sagemaker_xgboost_container/data_utils.py
edwardjkim: On autopilot side I assume? I would feel more comfortable if we have some tests on the xgboost side since there is significant code change.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -84,7 +86,13 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
     val_dmatrix = get_dmatrix(validate_path, content_type, is_pipe=is_pipe) \
         if val_files_size > 0 else None
 
-    return train_dmatrix, val_dmatrix
+    train_val_dmatrix = train_dmatrix
```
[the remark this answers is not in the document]
lihaife: I have been swinging between these two options. If for any reason we fail to combine the matrices, we won't do cross validation as train_val_dmatrix is None. If it is at least same as train_dmatrix, cross validation may still be performed even though the results is less optimal.

I am fine with either approach.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -124,14 +132,17 @@ def sagemaker_train(train_config, data_config, train_path, val_path, model_dir,
     is_pipe = (input_mode == Channel.PIPE_MODE)
 
     validation_channel = validated_data_config.get('validation', None)
-    train_dmatrix, val_dmatrix = get_validated_dmatrices(train_path, val_path, file_type, csv_weights, is_pipe)
+    combine_train_val = 'nfold' in validated_train_config
```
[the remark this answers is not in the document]
lihaife: if `fold < 2`, the hyper parameter validation will fail first. If you prefer, we may add more safe guard here.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -221,9 +233,18 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir,
         logging.info("Validation matrix has {} rows".format(val_dmatrix.num_row()))
 
     try:
+        nfold = train_cfg.pop("nfold", None)
```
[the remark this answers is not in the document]
lihaife: `nfold` is not an accepted parameter for `xgb.train()`. So pop it first before calling `xgb.train()`.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -221,9 +233,18 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir,
...
+
         bst = xgb.train(train_cfg, train_dmatrix, num_boost_round=num_round, evals=watchlist, feval=configured_feval,
                         early_stopping_rounds=early_stopping_rounds, callbacks=callbacks, xgb_model=xgb_model,
                         verbose_eval=False)
+
+        if nfold is not None and train_val_dmatrix is not None:
+            logging.info("Run {} fold cross validation on the data of {} rows".format(nfold,
+                                                                                      train_val_dmatrix.num_row()))
+            xgb.cv(train_cfg, train_val_dmatrix, nfold=nfold, num_boost_round=num_round,
+                   feval=configured_feval, early_stopping_rounds=early_stopping_rounds, show_stdv=True,
```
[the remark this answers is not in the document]
lihaife: Autopilot is doing research to enable early_stopping for XGB and consequently will make num_boost_round non-tunable. So early_stopping is important and future proof.

src/sagemaker_xgboost_container/data_utils.py
lihaife: DMatrix is an one-way door. The only way to combine two input channels, it is to read them together. In fact, mlio lib support to read two pipe channels together. We extend this function a little to leverage that capability.

I add the test in xgb's integration tests. I also did end-to-end tests with local runs and AP.

Because cross validation relies on this capability, I have to make the PR for both.

lihaife:
To get best results from cross validation, we should merge `train_dmatrix` and `val_dmatrix` for bigger data. However, `DMatrix` doesn't support concat operation and it cannot be exported to other formats (e.g. numpy). It is possible to write it to a file in binary format `matrix.save("data.buffer")`. However, xgb doesn't support read multiple buffer files.

So it is an one-way door. The only way to combine the data is to read them in one shot. Fortunately, milo supports multiple pipes together. So we extends `data_utils.get_dmatrix()` to support multiple files/pipes, when the parameter `data_path` is a list.

src/sagemaker_xgboost_container/data_utils.py
lihaife: The test is in test_data_utils.py. Scroll down to the bottom.

edwardjkim (approved):

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -221,9 +233,30 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir,
...
+            logging.info("Print the metrics of last epoch in the format expected by HPO")
+            cv_last_epoch = len(cv_eval_result.index) - 1
+            cv_eval_report = f"[{cv_last_epoch}]"
+            cv_eval_columns = cv_eval_result.columns
+            # Skip the standard deviation columns
+            for j in range(0, len(cv_eval_columns), 2):
+                metric_name = cv_eval_columns[j][:-5].replace("test-", "validation-", 1)
+                metric_val = cv_eval_result.at[cv_last_epoch, cv_eval_columns[j]]
+                cv_eval_report += '\t{0}:{1:.5f}'.format(metric_name, metric_val)
+            print(cv_eval_report)
```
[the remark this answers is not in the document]
lihaife: `logging.info()` will add the prefix such as `INFO:root:`. With plain `print(()`, we match the exact format as xgb prints the metrics.

PR #179, opened 2021-03-17
Cherry-pick commits from master to sync 0.90-1 branch with production image

edwardjkim:
*Description of changes:*

This PR is a cherry-pick of the commits in the master branch up to and including the hp validation eval metrics (#82). It was recently discovered that these commits have been deployed to production, but the 0.90-1 branch did not have these commits. Specifically, the new image built from https://github.com/aws/sagemaker-xgboost-container/commit/7d9a9c662215c515aeb9640846b45ac2be16587c did not pass the 0.90-1 integration tests.

*Testing*
0.90-1 integration tests passed.


src/sagemaker_xgboost_container/constants/xgb_constants.py
```diff
@@ -10,6 +10,31 @@
 # distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+
+XGB_MAXIMIZE_METRICS = [
```
aws-patlin, line 14: Some of these are actually mis-categorized. We should probably fix them in this PR.

According to master: https://github.com/aws/sagemaker-xgboost-container/blob/master/src/sagemaker_xgboost_container/constants/xgb_constants.py#L14

iyerr3: +1 - worthwhile pulling https://github.com/aws/sagemaker-xgboost-container/commit/cd4fb0d86a608c050e714c9125b01847f4bc4587 only if it's low cost and only if it's already in 0.90-2 (otherwise it will become a regression).

aws-patlin: It actually isn't in 0.90-2. It looks like that commit was only applied from 1.0-1 onward. Should we leave it as-is then, or also cherry-pick it into 0.90-2?

iyerr3: I suggest we leave it for now since we don't want to delay this fix.

aws-patlin (approved):

PR #189, opened 2021-03-25
Repeated cross validation and ensemble prediction

lihaife:
*Issue #, if available:*

*Description of changes:*

Support repeated cross validation and ensemble prediction with multiple cross validation models. Because `xgb.cv()` doesn't return models, we implement (repeated) cross validation with multiple `xgb.train()` call. As its behavior is different from `xgb.cv()` (e.g. early stopping may stop each fold at different round), we change the parameter name to `_kfold` and would like to reserve `nfold` (and `_nfold`) to the use cases that employs `xgb.cv()`.


iyerr3:
1. Can you clean up the commit list and organize so that it's easier to see the changes commit-to-commit? 
2. I'm having a hard time understanding the value of this. We're adding stacking/averaging on top of existing ensemble method? Do we have a scientific backing of this working better than just adding more complexity in an existing XGBoost model?

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -203,14 +201,100 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
+                            callbacks=callbacks, xgb_model=xgb_model, verbose_eval=False)
+
+        else:
+            logging.info("Run {}-fold cross validation with {} rows".format(kfold, train_val_dmatrix.num_row()))
+
+            bst = []
+            evals_results = []
+
+            num_cv_round = train_cfg.pop("_num_cv_round", 1)
+            for cv_round in range(num_cv_round):
```
iyerr3, line 228: this is difficult to follow - we need an explanation on what's a 'round' vs 'fold' and why we need both.

lihaife: The 'round' is about how many times we run cross validation. Therefore, it is known as repeated cross validation (we reshuffle data before each round). The purpose is to generate sufficient number of models for ensemble. The benchmarks show that ensemble of 3 rounds of cross validation models increase the accuracy significantly.

The fold refers the each training inside of one cross validation.

lihaife:
- Can you clean up the commit list and organize so that it's easier to see the changes commit-to-commit?

How to clean up the commit? I am not clear. If we do "squash and merge", all commits will be merged as one commit.

- I'm having a hard time understanding the value of this. We're adding stacking/averaging on top of existing ensemble method? Do we have a scientific backing of this working better than just adding more complexity in an existing XGBoost model?

The benchmarks have shown that ensemble with cross validation (especially repeated cross validation) improves the accuracy significantly.

src/sagemaker_xgboost_container/algorithm_mode/train.py
iyerr3: My concern is not the repeated cross-validation but rather using the correlated models to build a stack of XGBoost models for inference. I'm not convinced that is a good approach and am looking for scientific papers that provide the theory and results. I strongly believe that the accuracy you will achieve will be same if you increase complexity of base XGBoost model to match the model complexity introduced by combining multiple models.

iyerr3:
The repeated CV for model evaluation seems ok and that can be done with the in-built `xgb.cv()` function. It's the "ensembling" or "stacking" that I would like some backing (beyond the internal benchmarking).

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -203,14 +201,100 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
+                if num_cv_round > 1:
+                    logging.info("Run round {} cross validation".format(cv_round+1))
+
+                train_val_size = train_val_dmatrix.num_row()
+                row_index = np.random.permutation(range(train_val_size))
+                fold_size = round(train_val_size / kfold)
+                for fold in range(kfold):
```
iyerr3, line 235: Can we simplify by using sklearn function ([RepeatedKFold](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.RepeatedKFold.html)) for the indices?

lihaife: Good to know. Will update. Thanks.

lihaife:
[AutoGluon](https://arxiv.org/abs/2003.06505) does ensemble with repeated cross validation with XGBoost, lightGBM, RandomForest and a few other algorithms. Their paper shows significant accuracy improvement with this strategy even though XGBoost, lightGBM and RandomForest are all ensemble algorithm already.

src/sagemaker_xgboost_container/algorithm_mode/train.py
lihaife: Updated with `RepeatedKFold`.

src/sagemaker_xgboost_container/algorithm_mode/train.py
lihaife: [AutoGluon](https://arxiv.org/abs/2003.06505) does ensemble with repeated cross validation with XGBoost, lightGBM, RandomForest and a few other algorithms. Their paper shows significant accuracy improvement with this strategy even though XGBoost, lightGBM and RandomForest are all ensemble algorithm already.

edwardjkim:
The CI is failing due to [docker pull rate limit](https://www.docker.com/increase-rate-limits#:~:text=To%20increase%20your%20pull%20rate,accounts%20are%20now%20in%20effect.) because it uses an anonymous user to `docker pull`. We have to ultimately fix this, but for the short term, I would simply wait a few hours and push a dummy commit, e.g., `git commit --allow-empty -m "Trigger Build"` to make the CI succeed.

iyerr3 (approved):

PR #190, opened 2021-03-29
Add f1_binary and f1_macro

huibinshen:
Change `f1` to `f1macro` to make it more explicit on what it does. `f1macro` should be used for multi-class classification datasets.
For binary classification tasks, use `f1` (with the average option 'binary').

### DIFF
--- a/src/sagemaker_xgboost_container/constants/xgb_constants.py
+++ b/src/sagemaker_xgboost_container/constants/xgb_constants.py
@@ -16,6 +16,8 @@ XGB_MAXIMIZE_METRICS = [
     'auc',
     'aucpr',
     'f1',
+    'f1_binary',
+    'f1_macro',
     'map',
     'ndcg',
     'r2'
--- a/src/sagemaker_xgboost_container/constants/xgb_constants.py
+++ b/src/sagemaker_xgboost_container/constants/xgb_constants.py
@@ -44,6 +44,7 @@ XGB_MINIMIZE_METRICS = [
 
 LOGISTIC_REGRESSION_LABEL_RANGE_ERROR = "label must be in [0,1] for logistic regression"
 MULTI_CLASS_LABEL_RANGE_ERROR = "label must be in [0, num_class)"
+MULTI_CLASS_F1_BINARY_ERROR = "Target is multiclass but average='binary'"
 FEATURE_MISMATCH_ERROR = "feature_names mismatch"
 LABEL_PREDICTION_SIZE_MISMATCH = "Check failed: preds.size() == info.labels_.size()"
 ONLY_POS_OR_NEG_SAMPLES = "Check failed: !auc_error AUC: the dataset only contains pos or neg samples"
@@ -56,6 +57,7 @@ REG_LAMBDA_ERROR = "Parameter reg_lambda should be greater equal to 0"
 CUSTOMER_ERRORS = [
     LOGISTIC_REGRESSION_LABEL_RANGE_ERROR,
     MULTI_CLASS_LABEL_RANGE_ERROR,
+    MULTI_CLASS_F1_BINARY_ERROR,
     FEATURE_MISMATCH_ERROR,
     LABEL_PREDICTION_SIZE_MISMATCH,
     ONLY_POS_OR_NEG_SAMPLES,
--- a/src/sagemaker_xgboost_container/metrics/custom_metrics.py
+++ b/src/sagemaker_xgboost_container/metrics/custom_metrics.py
@@ -10,8 +10,6 @@
 # distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
-from collections import Counter
-
 import numpy as np
 from sklearn.metrics import accuracy_score, f1_score, mean_squared_error, r2_score
 
--- a/src/sagemaker_xgboost_container/metrics/custom_metrics.py
+++ b/src/sagemaker_xgboost_container/metrics/custom_metrics.py
@@ -86,7 +86,7 @@ def f1_binary(preds, dtrain):
 
 
 def f1_macro(preds, dtrain):
-    """Compute f1 score. This can be used for multiclassification training.
+    """Compute f1 macro score. This can be used for multiclassification training.
 
     For more information see: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html
 
--- a/src/sagemaker_xgboost_container/metrics/custom_metrics.py
+++ b/src/sagemaker_xgboost_container/metrics/custom_metrics.py
@@ -78,7 +78,6 @@ def f1_binary(preds, dtrain):
     if preds.size > 0:
         labels = dtrain.get_label()
         pred_labels = margin_to_class_label(preds)
-        # this function is used only for AutoPilot and the least frequent label is already encoded as 1.
         score = f1_score(labels, pred_labels, average='binary')
     return 'f1_binary', score
 
--- a/src/sagemaker_xgboost_container/metrics/custom_metrics.py
+++ b/src/sagemaker_xgboost_container/metrics/custom_metrics.py
@@ -10,6 +10,8 @@
 # distributed on an 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
+from collections import Counter
+
 import numpy as np
 from sklearn.metrics import accuracy_score, f1_score, mean_squared_error, r2_score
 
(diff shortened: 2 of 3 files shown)


src/sagemaker_xgboost_container/constants/xgb_constants.py
```diff
@@ -16,6 +16,7 @@
     'auc',
     'aucpr',
     'f1',
+    'f1macro',
```
huibinshen, line 19: Not completely sure about the string. Where I can see the possible metrics strings to be used for AutoPilot?

lihaife: It is used in hyperparameter settings, e.g. eval_metric or tuning metric, in format `f1` or `validation:f1`. XGB will strip off `validation:` part.

lihaife: Whenever this fix is accepted and merged, we should also update Autopilot pipeline recommender. In the file `algorithm/algo_metadata/config/sagemaker-xgboost/v0/sagemaker-xgboost.json`, update `F1` metric for multiclass to `validation:f1macro`. Thanks.

huibinshen: > It is used in hyperparameter settings, e.g. eval_metric or tuning metric, in format `f1` or `validation:f1`. XGB will strip off `validation:` part.

Thanks Haifeng. In pipeline recommender (algorithm/core/const.py), the metrics are `F1` and `F1_MACRO`, is it somewhere in this package, all the metrics names are converted to lower cases such that we could use 'f1' here? If this is true, should I use `f1_macro` instead of 'f1macro'?

huibinshen: > Whenever this fix is accepted and merged, we should also update Autopilot pipeline recommender. In the file `algorithm/algo_metadata/config/sagemaker-xgboost/v0/sagemaker-xgboost.json`, update `F1` metric for multiclass to `validation:f1macro`. Thanks.

I think we use accuracy as default metric for multi-class datasets in AutoPilot. User can override this choice to `f1macro` though.

lihaife: The metric name conversion is defined in `algorithm/algo_metadata/config/sagemaker-xgboost/v0/sagemaker-xgboost.json`. It is up to you to define it as `f1macro` or `f1_macro`.

Note that we define multiple metrics for each problem type. We have F1 and ACCURACY defined for multiclass. ACCURACY is the default one. It doesn't mean that you cannot define other metrics for multiclass.

iyerr3: I suggest `f1_macro` - easier to parse

huibinshen: Alright, thanks. Will change it to `f1_macro`.

huibinshen: @iyerr3 @lihaife How about XGBoost 1.0? Do we need to change it there as well?

lihaife: I don't suggest back port breaking changes.

src/sagemaker_xgboost_container/metrics/custom_metrics.py
```diff
@@ -51,6 +53,24 @@ def accuracy(preds, dtrain):
...
+    score = 0.0
+    if preds.size > 0:
+        labels = dtrain.get_label()
+        pred_labels = margin_to_class_label(preds)
+        pos_label = Counter(labels).most_common()[-1][0]  # using least frequent label as positive label
+        score = f1_score(labels, pred_labels, average='binary', pos_label=pos_label)
+    return 'f1', score
+
+
+def f1macro(preds, dtrain):
```
lihaife, line 73: Why not `f1_macro`? It follows the python code styling and also matches the metric name.

huibinshen: Make totally sense, will change it in the next revision.

edwardjkim:
The CI is failing due to [docker pull rate limit](https://www.docker.com/increase-rate-limits#:~:text=To%20increase%20your%20pull%20rate,accounts%20are%20now%20in%20effect.) because it uses an anonymous user to `docker pull`. We have to ultimately fix this, but for the short term, I would simply wait a few hours and push a dummy commit, e.g., `git commit --allow-empty -m "Trigger Build"` to make the CI succeed.

edwardjkim:
flake8 error:
```
src/sagemaker_xgboost_container/metrics/custom_metrics.py:13:1: F401 'collections.Counter' imported but unused
979
```

lihaife (approved):

huibinshen:
> flake8 error:
> 
> ```
> src/sagemaker_xgboost_container/metrics/custom_metrics.py:13:1: F401 'collections.Counter' imported but unused
> 979
> ```

huibinshen:
> flake8 error:
> 
> ```
> src/sagemaker_xgboost_container/metrics/custom_metrics.py:13:1: F401 'collections.Counter' imported but unused
> 979
> ```

Thanks, fixed.

edwardjkim:
Sorry, another docker pull rate limit error. Is the PR ready? I can start running integration tests if ready.

huibinshen:
> Sorry, another docker pull rate limit error. Is the PR ready other than flake8? I can start running integration tests if ready.

I can't see the failed tests, and not 100% sure if there is any other issues besides the pull rate limit error. I could trigger another dummy commit tomorrow morning to confirm.

huibinshen:
@iyerr3 Could you have a look of this change? Thanks.

src/sagemaker_xgboost_container/metrics/custom_metrics.py
```diff
@@ -52,6 +52,39 @@ def accuracy(preds, dtrain):
...
+
+    :param preds: Prediction values
+    :param dtrain: Training data with labels
+    :return: Metric name, f1 score
+    """
+    score = 0.0
+    if preds.size > 0:
+        labels = dtrain.get_label()
+        pred_labels = margin_to_class_label(preds)
+        # this function is used only for AutoPilot and the least frequent label is already encoded as 1.
```
iyerr3, line 81: There's nothing that stops external customers from using this directly, so I wouldn't make this assumption. 
Also I'm not sure what the implications of the 'least frequent label is already encoded as 1' is. Can you restate the assumption here to ensure behavior is clear. 

Also do we need validation for the number of labels to be `2`?

lihaife: Why does this assumption matter? For F1, we just care what's positive/negative.

huibinshen: To make f1 binary work, the only requirement is the label is encoded in {0, 1}. Taking least frequent label as 1 is only a behavior AutoPilot enforces. The customer could also use more frequent label as 1.

A question here is that are the labels already encoded in {0, 1}? If not, we need to have another label encoder to make f1 binary works.

lihaife: @huibinshen Note that you get the labels from `margin_to_class_label` function, which always encodes in `{0,1}`. So your code is fine. More importantly, XGBoost is a product itself. You shouldn't add comments about Autopilot's behavior here. Just remove the comment, which is the root cause of confusions.

huibinshen: Thanks Haifeng, will remove the comment in the next revision.

@iyerr3 Do you have other comments? I hope the explanation above is clear.

iyerr3: Thanks for removing the comment. 
My question on validation still stands: if customer uses `f1_binary` with multi-class, then they should get a Customer Error. Is that validation happening somewhere that I missed?

huibinshen: Hi Rahul, I added a customer error in the new revision.

src/sagemaker_xgboost_container/constants/xgb_constants.py
```diff
@@ -42,6 +44,7 @@
 
 LOGISTIC_REGRESSION_LABEL_RANGE_ERROR = "label must be in [0,1] for logistic regression"
 MULTI_CLASS_LABEL_RANGE_ERROR = "label must be in [0, num_class)"
+MULTI_CLASS_F1_BINARY_ERROR = "Target is multiclass but average='binary'"
```
lihaife, line 47: Why do we throw this error?

huibinshen: When user uses F1 binary score for multiclass dataset, f1_sccore(..., average="binary") will throw this error. This is because a mis-configuration, so it is now thrown as a CustomerError.

edwardjkim (approved):
Changes including customer error look good to me. Integration tests all succeeded.

PR #195, opened 2021-04-26
Early stopping

lihaife:
*Issue #, if available:*

*Description of changes:*

Support the early stopping feature based on new 1.3 callback interface.


lihaife:
This is early review intended for 1.3 release.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -299,19 +309,26 @@ def get_callbacks_watchlist(train_cfg, train_dmatrix, val_dmatrix, model_dir, ch
...
     # Parse arguments for intermediate model callback
     save_model_on_termination = train_cfg.pop('save_model_on_termination', "false")
     if save_model_on_termination == "true":
         model_name = f"{MODEL_NAME}-{fold}" if fold is not None else MODEL_NAME
-        save_intermediate_model = checkpointing.save_intermediate_model(model_dir, model_name)
-        callbacks.append(save_intermediate_model)
+        callbacks.append(checkpointing.SaveIntermediateModelCallBack(model_dir, model_name))
         add_sigterm_handler(model_dir, is_master)
 
+    if early_stopping_data and early_stopping_metric and early_stopping_rounds:
```
huibinshen, line 323: If val_dmatrix is None, is it better to skip early stopping?

lihaife: It is very dangerous to early stop on the training data. I believe that MLP has the same design. Without validation data, MLP won't do early stopping.

huibinshen: I agree. Now if early_stopping_data == 'train', the early stopping will still be used, no?

lihaife: yes. But `early_stopping_data` is not a customer parameter (it is even not a string). We set it to `val_matrix` or `None`

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -189,19 +189,27 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
-    cleaned_eval_metric, configured_feval = train_utils.get_eval_metrics_and_feval(
+    cleaned_eval_metric, configured_feval, tuning_objective_metric = train_utils.get_eval_metrics_and_feval(
         tuning_objective_metric_param, eval_metric)
     if cleaned_eval_metric:
         train_cfg['eval_metric'] = cleaned_eval_metric
     else:
         train_cfg.pop('eval_metric', None)
 
+    early_stopping_rounds = train_cfg.pop('early_stopping_rounds', None)
+    early_stopping_data = 'validation' if val_dmatrix else 'train'
```
huibinshen, line 205: Hi Haifeng, see this line for `early_stopping_data`.

src/sagemaker_xgboost_container/checkpointing.py
```diff
@@ -388,4 +388,16 @@ def callback(self, env):
...
+class SaveIntermediateModelCallBack(xgb.callback.TrainingCallback):
+    """The new implementation of callback functions from 1.3."""
+    def __init__(self, intermediate_model_dir, model_name, is_master):
+        self.callback = SaveIntermediateModel(intermediate_model_dir, model_name)
+        self.is_master = is_master
+
+    def after_iteration(self, model, epoch, evals_log) -> bool:
+        if self.is_master:
+            self.callback.save_intermediate_model(model)
+        return False
```
aws-patlin, line 403: Is it right for this to always return False?

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -296,19 +308,28 @@ def get_callbacks_watchlist(train_cfg, train_dmatrix, val_dmatrix, model_dir, ch
         logging.info("Resuming from iteration %s", iteration)
 
     callbacks = []
-    callbacks.append(checkpointing.print_checkpointed_evaluation(start_iteration=iteration))
```
aws-patlin, line 299: Should we remove the `print_checkpointed_evaluation` function and its related functions/classes if they're no longer being used? Same thing for `save_checkpoint` and `save_intermediate_model`.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -188,19 +188,29 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
         train_cfg['eval_metric'] = cleaned_eval_metric
     else:
         train_cfg.pop('eval_metric', None)
 
+    early_stopping_rounds = train_cfg.pop('early_stopping_rounds', None)
+    early_stopping_data = 'validation' if val_dmatrix else 'train'
+    early_stopping_metric = None
+    if early_stopping_rounds:
+        if tuning_objective_metric:
+            early_stopping_metric = tuning_objective_metric[-1]
```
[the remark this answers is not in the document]
aws-patlin: And what's the reasoning for choosing the last metric to use for early stopping?

aws-patlin:
Looks like the CI failed--please take a look at the logs and fix the failure(s).

src/sagemaker_xgboost_container/algorithm_mode/train.py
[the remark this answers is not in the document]
lihaife: Both eval_metric and tuning_objective_metric have been there for long time. I agree that they are kind of misleading. But it is XGB's naming convention. I prefer to keep it as.

It is also the convention to choose the last metric as early stopping metric before. So we keep it as is so that we don't break customer's code.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -188,19 +188,29 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
-    cleaned_eval_metric, configured_feval = train_utils.get_eval_metrics_and_feval(
+    cleaned_eval_metric, configured_feval, tuning_objective_metric = train_utils.get_eval_metrics_and_feval(
         tuning_objective_metric_param, eval_metric)
     if cleaned_eval_metric:
         train_cfg['eval_metric'] = cleaned_eval_metric
     else:
         train_cfg.pop('eval_metric', None)
 
+    early_stopping_rounds = train_cfg.pop('early_stopping_rounds', None)
+    early_stopping_data = 'validation' if val_dmatrix else 'train'
```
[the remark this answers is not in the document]
lihaife: okay.

src/sagemaker_xgboost_container/checkpointing.py
lihaife: Yes. If we turn True, it means to stop the training. Most callbacks always return False.

src/sagemaker_xgboost_container/algorithm_mode/train.py
lihaife: They could be used in script mode. Note that they are in the top level module, not just in algorithm mode. So it is better to keep for now.

src/sagemaker_xgboost_container/checkpointing.py
lihaife: early_stopping_data is a function parameter here. Although we always set the value in the caller, it is still better to do the sanity check here.

src/sagemaker_xgboost_container/algorithm_mode/train.py
lihaife: fixed

src/sagemaker_xgboost_container/algorithm_mode/train.py
lihaife: fixed

aws-patlin (approved):

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -283,7 +293,9 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                 logging.debug("Stored trained model {} at {}".format(fold, model_location))
 
 
-def get_callbacks_watchlist(train_cfg, train_dmatrix, val_dmatrix, model_dir, checkpoint_dir, is_master, fold=None):
+def get_callbacks_watchlist(train_dmatrix, val_dmatrix, model_dir, checkpoint_dir,
```
awsbmillare, line 296: This arg list is getting pretty long, do you think we can turn this sig into named args?

lihaife: done

awsbmillare (approved):

PR #209, opened 2021-08-10
Latest multiarch

awsbmillare:
* Fix conda version bump
* Add arm64 dockerfile for arm support and eventual multiarch support

### DIFF
--- a/docker/1.3-1/base/Dockerfile_arm64.cpu
+++ b/docker/1.3-1/base/Dockerfile_arm64.cpu
@@ -1,0 +1,105 @@
+ARG UBUNTU_VERSION=18.04
+ARG IMAGE_DIGEST=c60266b67f58fafc30703315f617a8fcccaffc48ef5534ca5f67a9ba3aceb3b8
+
+FROM ubuntu:18.04
+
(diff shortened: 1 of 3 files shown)


docker/1.3-1/base/Dockerfile_arm64.cpu
```diff
@@ -0,0 +1,105 @@
+ARG UBUNTU_VERSION=18.04
+ARG IMAGE_DIGEST=c60266b67f58fafc30703315f617a8fcccaffc48ef5534ca5f67a9ba3aceb3b8
+
+FROM ubuntu:18.04
```
aws-patlin, line 4: No cuda support on arm64?

aws-patlin (approved):
How did you test the new dockerfile?

docker/1.3-1/base/Dockerfile_arm64.cpu
awsbmillare: Nada. Current Graviton instances don't have NVidia GPUs anyways so not an issue for SageMaker customers but Graviton3 will have GPU support so this may change in the future.

aws-patlin (approved):

PR #212, opened 2021-08-23
Remove hardcoded payload limit

awsbmillare:
*Issue #, if available:*

Even though we have a user configurable MAX_CONTENT_LENGTH is setting, we still enforce a hardcoded upper limit. This is no longer needed because the original requester of the limit no longer has this requirement and is actually asking for this limit to be removed as they have their own checking proceses.

*Description of changes:*

Remove upper limit on PARSED_MAX_CONTENT_LENGTH. That variable is is normally set by env var MAX_CONTENT_LENGTH but was restricted to under 20MB before, now there is no limit. (A default is still retained).


aws-patlin (approved):

awsbmillare:
codebuild hook had an issue that I resolved offline, confirmed that it passed though

PR #223, opened 2021-10-19
Fix non-alg mode content-type handling

awsbmillare:
*Issue #, if available:*

Internal high escalations

*Description of changes:*

Content-type handling of default_input_fn (for non-algorithm mode) currently has a brittle implementation that does not accept more sophisticated content-type formats. Content formats are well defined https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type and can have parameters after a semicolon. Currently, this container can support loading of csv files, where it will convert csv files into an xgboost dmatrix as long as the supplied content-type is 'text/csv'. However, it will throw an UnsupportedFormatError if passed 'text/csv; charset=UTF-8', which is incorrect as this should be supported. This change correctly accepts this case and opens the door to parsing/handling those additional parameters if needed in the future. Note that in the case of csv and UTF-8, python3 already handles the UTF-8 case with no other changes needed.

### DIFF
--- a/test/unit/test_encoder.py
+++ b/test/unit/test_encoder.py
@@ -99,8 +99,8 @@ def test_decode(content_type):
 
 @pytest.mark.parametrize('content_type', ['text/csv; charset=UTF-8'])
 def test_decode_with_complex_csv_content_type(content_type):
-    encoder.decode("42.0,6.0,9.0\n42.0,6.0,9.0", content_type)
-    assert type(actual) is xgb.DMatrix
+    dmatrix_result = encoder.decode("42.0,6.0,9.0\n42.0,6.0,9.0", content_type)
+    assert type(dmatrix_result) is xgb.DMatrix
 
 
 def test_encoder_jsonlines_from_json():


awsbmillare:
I can't link internal tickets and integ tests here but this is likely causing breakage right now. Also for non-alg we are NOT handling semicolons https://github.com/aws/sagemaker-xgboost-container/blob/master/src/sagemaker_xgboost_container/handler_service.py#L61 . For alg-mode we are https://github.com/aws/sagemaker-xgboost-container/blob/196bc874127101951365dc5d39e10bf3db494c61/src/sagemaker_xgboost_container/algorithm_mode/handler_service.py#L68, there are two different implementations here.

Currently my understanding is some platform/reverse-proxy is now calling the server with charset=utf-8 supplied, which was not happening in the past. This is currently the relief fix once this is confirmed to actually address the situation.

Just to compare to non-alg mode, that accepts content-type with multiple variants and ignores irrelevant params (and also accepts specific non-media-type valid forms as well). In these proposed changes, we simply ignore additional params. This increases what we accept on and in the future, if we care about throwing errors for other invalid params combinations, this would be a "breaking" change, but I don't see this being much an issue when it comes to "breaking" changes for clearly unspecified/unused data.

PR #229, opened 2021-12-08
renaming duplicate Symlinks to avoid collision on same path and some code reformatting

NikhilRaverkar:
FileExistsError when both train and validation dataset contain files with the same name

made the data_path in for loop for creating symlink to set(data_path). This way we will not try to create symlinks to the existing path and the error will not occur.

TT# https://t.corp.amazon.com/P55999366/overview


src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -530,7 +530,7 @@ def get_dmatrix(data_path, content_type, csv_weights=0, is_pipe=False):
             files_path = "/tmp/sagemaker_xgboost_input_data"
             shutil.rmtree(files_path, ignore_errors=True)
             os.mkdir(files_path)
-            for path in data_path:
```
mabunday, line 533: `data_path` is used as input in the logic branches above as well. Should they also take in a deduped `data_path`?

mabunday:
The `get_dmatrix` is quite large and encompasses multiple functions. Consider 

1. Renaming `get_files_path` to `get_files_path_from_string`
2. Defining a new type `DataPath = Union[List[str], str]`
2. Defining a new function `get_files_path` 
```python
def get_files_path(data_path: DataPath, is_pipe: bool) -> Optional[List[str]]:
    if is_pipe:
        return _get_pipe_mode_files_path(data_path)
    else:
        return _get_file_mode_files_path(data_path)
```
3. Define a new function `_get_pipe_mode_files_path`
```python
def _get_pipe_mode_files_path(data_path: DataPath) -> Optional[List[str]]:
    if isinstance(data_path, list):
        files_path = data_path
    else:
        files_path = [data_path]
        if not os.path.exists(data_path + '_0'):
            logging.info('Pipe path {} does not exist!'.format(data_path))
            return None
    return files_path
```
4. Defining a new function
```python
def _get_file_mode_files_path(data_path: DataPath) -> Optional[List[str]]:
    if not isinstance(data_path, list):
        if not os.path.exists(data_path):
            logging.info('File path {} does not exist!'.format(data_path))
            return None
        files_path = get_files_path(data_path)
    else:
        # Create a directory with symlinks to input files.
        files_path = "/tmp/sagemaker_xgboost_input_data"
        shutil.rmtree(files_path, ignore_errors=True)
        os.mkdir(files_path)
        for path in set(data_path):
            if not os.path.exists(path):
                return None
            if os.path.isfile(path):
                os.symlink(path, os.path.join(files_path, os.path.basename(path)))
            else:
                for file in os.scandir(path):
                    os.symlink(file, os.path.join(files_path, file.name))
     return files_path
```
5. Adding unit tests for the above three functions

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: This change is not working so going to change the logic.

awsbmillare:
I think this is worth finishing exploring why this fails but now I'm thinking again if it might be better to not create the double sized dmatrix when train and validate are the same since for the purpose of cross validation, adding more entries doesn't add any value. At the very least we should emit a warning to customer that this is a strange thing to do. The customer when specifying _kfold, will trigger combining these together. I feel we should check if train and validate are the same and emit a warning. We need to make it clear to the customer that during k_fold training and validation data are combined. I'm not sure if that's clear yet.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -531,13 +531,17 @@ def get_dmatrix(data_path, content_type, csv_weights=0, is_pipe=False):
             shutil.rmtree(files_path, ignore_errors=True)
             os.mkdir(files_path)
             for path in data_path:
+                # logging for debugging remove before final merge
+                print(path)
```
awsbmillare, line 535: did you get the testing results for this yet?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -531,13 +531,19 @@ def get_dmatrix(data_path, content_type, csv_weights=0, is_pipe=False):
...
-                if os.path.isfile(path):
-                    os.symlink(path, os.path.join(files_path, os.path.basename(path)))
+                symlink_dest = os.path.join(files_path, os.path.basename(path))
+                if os.path.isfile(path) and not os.path.exists(symlink_dest):
+                    os.symlink(path, symlink_dest)
                 else:
+
                     for file in os.scandir(path):
-                        os.symlink(file, os.path.join(files_path, file.name))
+                        symlink_dest_dir = os.path.join(files_path, file.name)
```
awsbmillare, line 544: we need to test here how using 1 file instead of 2 when things collide if this will cause different expected results, and either emit warning or make the names different so that they don't collide. If we make new names, we need to make sure those new file names are used somehow

NikhilRaverkar:
### Description:

* For an input with same file name, XGB was trying to create symlink with same name which was resulting in error on customer side 
* Appending extra index at the end of file name in file mode to avoid collision caused by duplicate input file name
* Appending an index makes file name unique and it create different symlinks and is able to red both the files
* Moved Pipe mode and file mode `file_path` generation to different functions and added unit tests for them
* Added a couple of debug statements 

### Testing:
* Unit tests added
* Tested on local jupyter instance 

### Issues:
* https://t.corp.amazon.com/P55999366/overview

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +483,58 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
...
+    # directories to meet XGB's assumption that all files are in the same directory.
+
+    if isinstance(data_path, list):
+        logging.info('File path {} of input files'.format(data_path))
+        # Create a directory with symlinks to input files.
+        files_path = "/tmp/sagemaker_xgboost_input_data"
+        shutil.rmtree(files_path, ignore_errors=True)
+        os.mkdir(files_path)
+        for index, path in enumerate(data_path):
+            # logging for debugging remove before final merge
```
awsbmillare, line 515: are you going to clean this?

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -90,6 +90,9 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
     train_val_dmatrix = train_dmatrix
     if combine_train_val and train_dmatrix is not None and val_dmatrix is not None:
         logging.info("Read both train and validation data into one DMatrix")
+        if train_path == validate_path:
```
awsbmillare, line 93: Have you also tested the case where the paths might be different but the filename is the same and thus in the symlink creation would still create a collision?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +483,58 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
...
+                return None
+            if os.path.isfile(path):
+                base_name = os.path.join(files_path, os.path.basename(path) + '_' + str(index))
+                logging.info('creating symlink between Path {} and destination {}'.format(path, base_name))
+                os.symlink(path, base_name)
+            else:
+                for file in os.scandir(path):
+                    base_name = os.path.join(files_path, file.name + '_' + str(index))
+                    logging.info('creating symlink between Path {} and destination {}'.format(file, base_name))
+                    os.symlink(file, base_name)
```
awsbmillare, line 527: nit: you can factor out this duplicate code with an internal helper fun
args required are only source_path and name. Something like
```
def _make_symlink(source_path, name):
    base_name = os.path.join(files_path, name + '_' + str(index))
    logging.info('creating symlink between Path {} and destination {}'.format(source_path, base_name))
    os.symlink(source_path, base_name)
```

and then in the if bodies, you just call:
```
if os.path.isfile(path):
    _make_symlink(path, os.path.basename(path))
else:
    for file in os.scandir(path):
        _make_symlink(file, file.name)
```

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: removed the comment

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: refactored

NikhilRaverkar:
Closing because PR was merged [with some additional changes](https://github.com/aws/sagemaker-xgboost-container/pull/285)

PR #230, opened 2021-12-10
Store ground truth and predictions on validation folds to output folder

iaroslav-ai:
*Description of changes:*
Stores predictions of xgboost model on cross validation folds to output folder, so that those predictions can be used for any postprocessing. 

Our use case is to calculate confusion matrix and a few other statistics for model produced by this container; This requires predictions data that is not used for training. Currently this container with cv enabled loads all data in memory (training + validation) and uses both partitions for training of model ensemble. As we only have training / validation partition (and cannot easily add extra test partition) this change is storing predictions similar to https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_predict.html, which is the only way for us to be able to calculate correctly performance metrics after the container has terminated. 

To store additional outputs, the feature of training platform to output additional arguments is used: https://github.com/aws/sagemaker-containers#sm-output-data-dir . 

For output folder, the data is written into “data” subdirectory following convention here: https://github.com/aws/sagemaker-training-toolkit/blob/master/src/sagemaker_training/environment.py#L136 .

Testing performed: 
* Unit tests passing 
* Integration tests passing 
* Run xgboost training on ~166 benchmarking datasets 


src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -265,15 +282,48 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
+                aggregated_data_to_store = stats.mode(data_to_store, axis=1).mode[:, 0, :]
+                aggregated_data_to_store[:, prob_idx] = proba_avg
+            else:
+                aggregated_data_to_store = data_to_store.mean(axis=1)
+
+            if not os.path.exists(additional_output_path):
+                os.makedirs(additional_output_path)
+
+            pred_path = os.path.join(additional_output_path, 'predictions.csv')
+            logging.info(f"Storing predictions on cv folds averaged over all cv rounds in {pred_path} ")
```
mabunday, line 320:
```suggestion
            logging.info(f"Storing predictions on cv folds averaged over all cv rounds in {pred_path}")
```

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -242,14 +246,27 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
+            # different cross validation repeats - averaged for regression, mode is calculated for classification.
+            data_to_store = np.zeros((num_rows_in_dataset, num_cv_round, 3 if classification_problem else 2))
+            truth_idx = 0
+            prob_idx = 1
+            pred_idx = -1
+
+            # This vector is needed to average over repeates of k-fold cross validaiton.
+            cv_repeat_idx = np.zeros((num_rows_in_dataset,)).astype(int)
+
+            for train_index, val_idx in rkf.split(X=X, y=y):
```
mabunday, line 267: Really minor, but since `val_index` was renamed to `val_idx` and the data store indices use the suffix `_idx` as well, let's rename `train_index` to `train_idx` as well.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -265,15 +282,48 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
+                cv_repeat_idx[val_idx] += 1
                 evals_results.append(evals_result)
 
                 if len(bst) % kfold == 0:
                     logging.info("The metrics of round {} cross validation".format(int(len(bst) / kfold)))
                     print_cv_metric(num_round, evals_results[-kfold:])
 
+            if classification_problem:
+                proba_avg = data_to_store[:, :, prob_idx].mean(axis=1)
+                # mode always return same dimension of output
```
mabunday, line 310:
```suggestion
                # mode always returns same dimension of output
```

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -265,15 +282,48 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                                     evals=watchlist, feval=configured_feval, evals_result=evals_result,
                                     callbacks=callbacks, xgb_model=xgb_model, verbose_eval=False)
                 bst.append(booster)
+
+                # store predictions on validation fold
+                data_to_store[val_idx, cv_repeat_idx[val_idx], truth_idx] = cv_val_dmatrix.get_label()
+                if classification_problem:
+                    probabilities = booster.predict(cv_val_dmatrix)
+                    if probabilities.ndim == 2:
+                        # multi - class classification setting
```
mabunday, line 291:
```suggestion
                        # multiclass classification setting
```

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -242,14 +246,27 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
+            X = range(num_rows_in_dataset)
+            y = train_val_dmatrix.get_label() if classification_problem else None
             rkf = RepeatedStratifiedKFold(n_splits=kfold, n_repeats=num_cv_round) if y is not None \
                 else RepeatedKFold(n_splits=kfold, n_repeats=num_cv_round)
 
-            for train_index, val_index in rkf.split(X=X, y=y):
+            # For regression, ground truth and predicted values are stored. For classification, additionally
+            # estimated probability for predicted class is stored. Predictions are aggregated across
+            # different cross validation repeats - averaged for regression, mode is calculated for classification.
+            data_to_store = np.zeros((num_rows_in_dataset, num_cv_round, 3 if classification_problem else 2))
```
mabunday, line 259: Would naming this `postprocessing_data` be accurate? Just to be explicit that this data includes data that is not used for training.

mabunday (approved):
Looks good to me aside from some very minor comments.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -232,6 +234,8 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
 
         else:
             num_cv_round = train_cfg.pop("_num_cv_round", 1)
+            additional_output_path = os.environ[SM_OUTPUT_DATA_DIR]
```
awsbmillare, line 237: have you confirmed if this ENV var is always set? Might be a good idea to fallback to a sane default here. Alternatively, we'd need to check later if features are required and this is not set, then throw error there.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -242,14 +246,27 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
-            y = train_val_dmatrix.get_label() if num_class or objective.startswith("binary:") else None
+            classification_problem = num_class or objective.startswith("binary:")
+            num_rows_in_dataset = train_val_dmatrix.num_row()
+            X = range(num_rows_in_dataset)
+            y = train_val_dmatrix.get_label() if classification_problem else None
             rkf = RepeatedStratifiedKFold(n_splits=kfold, n_repeats=num_cv_round) if y is not None \
                 else RepeatedKFold(n_splits=kfold, n_repeats=num_cv_round)
 
-            for train_index, val_index in rkf.split(X=X, y=y):
+            # For regression, ground truth and predicted values are stored. For classification, additionally
```
awsbmillare, line 256: Can we call out here in the comments that the next block is for initializing variables needed for storing postprocessing data. This train_job method is already large and needs to be broken up. For now let's be clear here to make it easier in the future to separate things out. I want to scope out what's needed for particular parts of this method.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -265,15 +282,48 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                                     evals=watchlist, feval=configured_feval, evals_result=evals_result,
                                     callbacks=callbacks, xgb_model=xgb_model, verbose_eval=False)
                 bst.append(booster)
+
```
awsbmillare, line 285: This code seems isolated to only updating post-processing intermediate data. To minimize indentation growth and limit scope, can we factor this out into a helper function _train_job_update_postprocessing_data or something of that name. Let's prefer named args for the helper just like in the calls above.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -265,15 +282,48 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
+                    data_to_store[val_idx, cv_repeat_idx[val_idx], pred_idx] = booster.predict(cv_val_dmatrix)
+
+                cv_repeat_idx[val_idx] += 1
                 evals_results.append(evals_result)
 
                 if len(bst) % kfold == 0:
                     logging.info("The metrics of round {} cross validation".format(int(len(bst) / kfold)))
                     print_cv_metric(num_round, evals_results[-kfold:])
 
+            if classification_problem:
```
awsbmillare, line 308: similarly, for the "closing/writing" step, can we also factor this out into a helper function like _train_job_save_postprocessing_data?

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: Yup that is more accurate

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: I agree. 

I am thinking to have a class that is responsible for aggregation of predictions, and storing those predictions to file. This way it is easier to extend to e.g. make predictions on non cv folds, easier to copy to other algorithm containers, and most of the new code would be factored in.

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: +1 will refactor

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: Updated

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: Updated

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: Updated

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: I think I removed this comment

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: Refactored to a dedicated class

iaroslav-ai:
Thanks all for your comments! I refactored the code for storage of the predictions on validation folds into its own class. Unittests are missing for the functionality of this class, but integration tests are passing. I will add unittests if we are ok with this new class.

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def save(self):
+        """Persists to file system predictions in user specified path."""
+        columns = [self.y_true]
+        if self.classification:
+            columns.append(self.y_prob.mean(axis=-1))
+            # mode always returns same number of dimensions of output as for input
+            columns.append(stats.mode(self.y_pred, axis=1).mode[:, 0])
+        else:
+            columns.append(self.y_pred.mean(axis=-1))
```
mabunday, line 71: Let's split this into a helper function and call it directly in the call to `np.savetxt`.

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+import logging
+import numpy as np
+from scipy import stats
+
+PREDICTIONS_OUTPUT_FILE = 'predictions.csv'
+
+
+class ValidationPredictionRecorder:
+    """Helper class to record and store predictions obtained on different train / validation
+    folds. Predictions are stored in additional artefact folder, and as a result are stored
```
mabunday, line 23:
```suggestion
    folds. Predictions are stored in additional artifact folder, and as a result are stored
```

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+            else:
+                labels = 1 * (predictions > 0.5)
+                proba = predictions
+            self.y_pred[indices, cv_repeat_idx] = labels
+            self.y_prob[indices, cv_repeat_idx] = proba
+        else:
+            self.y_pred[indices, cv_repeat_idx] = predictions
+        self.cv_repeat_counter[indices] += 1
+
+    def save(self):
```
mabunday, line 63:
```suggestion
    def save(self) -> None:
```

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, y_true: np.ndarray, num_cv_round: int, classification: bool, output_data_dir: str):
+        self.y_true = y_true.copy()
+        num_rows = len(y_true)
+        self.y_pred = np.zeros((num_rows, num_cv_round))
+        self.y_prob = self.y_pred.copy() if classification else None
+        self.cv_repeat_counter = np.zeros((num_rows,)).astype(int)
+        self.classification = classification
+        self.output_data_dir = output_data_dir
+
+    def record(self, indices: np.ndarray, predictions: np.ndarray):
```
mabunday, line 40:
```suggestion
    def record(self, indices: np.ndarray, predictions: np.ndarray) -> None:
```

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    """Helper class to record and store predictions obtained on different train / validation
+    folds. Predictions are stored in additional artefact folder, and as a result are stored
+    in output path of training job as output.tar.gz.
+
+    Attributes:
+        y_true           (1d numpy array): Ground truth labels.
+        num_cv_round     (int): number times cross validation procedure will be repeated.
+        classification   (bool): indicates type of learning problem.
+    """
+    def __init__(self, y_true: np.ndarray, num_cv_round: int, classification: bool, output_data_dir: str):
```
mabunday, line 31:
```suggestion
    def __init__(self, y_true: np.ndarray, num_cv_round: int, classification: bool, output_data_dir: str) -> None:
```

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        self.y_prob = self.y_pred.copy() if classification else None
+        self.cv_repeat_counter = np.zeros((num_rows,)).astype(int)
+        self.classification = classification
+        self.output_data_dir = output_data_dir
+
+    def record(self, indices: np.ndarray, predictions: np.ndarray):
+        """Records in memory predictions on single validation fold.
+
+        If current host is master host, initialize and start the Rabit Tracker in the background. All hosts then connect
+        to the master host to set up Rabit rank.
```
mabunday, line 44: This looks like it belongs in the training loop instead.

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        self.y_true = y_true.copy()
+        num_rows = len(y_true)
+        self.y_pred = np.zeros((num_rows, num_cv_round))
+        self.y_prob = self.y_pred.copy() if classification else None
+        self.cv_repeat_counter = np.zeros((num_rows,)).astype(int)
+        self.classification = classification
+        self.output_data_dir = output_data_dir
+
+    def record(self, indices: np.ndarray, predictions: np.ndarray):
+        """Records in memory predictions on single validation fold.
```
mabunday, line 41:
```suggestion
        """Record predictions on a single validation fold in-memory.
```

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+                labels = 1 * (predictions > 0.5)
+                proba = predictions
+            self.y_pred[indices, cv_repeat_idx] = labels
+            self.y_prob[indices, cv_repeat_idx] = proba
+        else:
+            self.y_pred[indices, cv_repeat_idx] = predictions
+        self.cv_repeat_counter[indices] += 1
+
+    def save(self):
+        """Persists to file system predictions in user specified path."""
```
mabunday, line 64:
```suggestion
        """Serialize predictions to this instance's output data directory."""
```

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        columns = [self.y_true]
+        if self.classification:
+            columns.append(self.y_prob.mean(axis=-1))
+            # mode always returns same number of dimensions of output as for input
+            columns.append(stats.mode(self.y_pred, axis=1).mode[:, 0])
+        else:
+            columns.append(self.y_pred.mean(axis=-1))
+
+        if not os.path.exists(self.output_data_dir):
+            os.makedirs(self.output_data_dir)
```
mabunday, line 74: Let's also split this into a helper function since and add additional logging (e.g. "Path does not exist, creating...") since writing to the filesystem is a common failure path.

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,78 @@
...
+        :param predictions: predictions for rows specified in `indices` variable.
+        """
+        cv_repeat_idx = self.cv_repeat_counter[indices]
+        if self.classification:
+            if predictions.ndim > 1:
+                labels = np.argmax(predictions, axis=-1)
+                proba = predictions[np.arange(len(labels)), labels]
+            else:
+                labels = 1 * (predictions > 0.5)
+                proba = predictions
```
mabunday, line 56: Prediction dimension should remain constant between repeated calls to `record` for the same class instance, right? If so, once the first prediction is recorded, let's "fix" this logic and do some basic error handling/logging if the dimensions change.

mabunday (changes requested):

src/sagemaker_xgboost_container/algorithm_mode/train.py
iaroslav-ai: I think this should be always set by the training platform `https://github.com/aws/sagemaker-containers#sm-output-data-dir`

iaroslav-ai: I will double check

src/sagemaker_xgboost_container/prediction_utils.py
iaroslav-ai: Agreed. Added

src/sagemaker_xgboost_container/prediction_utils.py
iaroslav-ai: Added

src/sagemaker_xgboost_container/prediction_utils.py
iaroslav-ai: Good catch - this was a bad copy paste :)

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,107 @@
...
+    def record(self, indices: np.ndarray, predictions: np.ndarray) -> None:
+        """Record predictions on a single validation fold in-memory.
+
+        :param indices: indicates for which rows the predictions were made.
+        :param predictions: predictions for rows specified in `indices` variable.
+        """
+        if self.pred_ndim_ is None:
+            self.pred_ndim_ = predictions.ndim
+        if self.pred_ndim_ != predictions.ndim:
+            raise exc.AlgorithmError(
```
mabunday, line 54: First, thank you for all of the changes! 

Only one comment now: I think this and the other AlgorithmErrors below could be raised as `CustomerErrors` instead because they're issues with the input data, no? In that case it would be actionable by them but not actionable by us.

src/sagemaker_xgboost_container/prediction_utils.py
```diff
@@ -0,0 +1,107 @@
...
+                f"Expected predictions with ndim={self.pred_ndim_}, got ndim={predictions.ndim}."
+            )
+
+        cv_repeat_idx = self.cv_repeat_counter[indices]
+        if np.any(cv_repeat_idx == self.num_cv_round):
+            raise exc.AlgorithmError(
+                f"More than {self.num_cv_round} repeated predictions for same row were provided."
```
mabunday, line 61: Can we explicitly say which row as part of the message? Same below.

mabunday (changes requested):

mabunday:
Minor comments but looks good to me overall. Thanks for the changed!

src/sagemaker_xgboost_container/prediction_utils.py
iaroslav-ai: Currently this class is used only in xgboost algorithm mode. There should be no way for the user to provide data that causes xgboost estimator to return predictions of different dimensions, or fail other checks that I added, and I expect that for other use cases it should be the same. Because of that I would keep it as algorithm error

mabunday (approved):

src/sagemaker_xgboost_container/prediction_utils.py
iaroslav-ai: Will add

mabunday (approved):

test/integration/local/test_kfold.py
```diff
@@ -0,0 +1,64 @@
...
+            {
+                'objective': 'multi:softprob', 'num_class': '4', '_kfold': '5', '_num_cv_round': '3'
+            },
+            15
+        ),
+    ]
+)
+def test_xgboost_abalone_kfold(dataset, extra_hps, model_file_count, docker_image, opt_ml):
+    hyperparameters = get_abalone_default_hyperparameters()
+    data_path = os.path.join(path, "..", "..", "resources", dataset, "data")
```
awsbmillare, line 51: nit: can we reuse data_root above here?

awsbmillare (approved):
I have only a very minor nit, otherwise LGTM. Really nice work, thanks for all the changes.

test/integration/local/test_kfold.py
iaroslav-ai: Fixed, thanks!

PR #253, opened 2022-03-11
update version to 1.5.2 && resolve feature name breaking

haixiw:
*Issue #, if available:*

*Description of changes:*
There is a breaking change between Xgboost 1.3.3 to version 1.5.2:

XGBoost used to generate some pseudo feature names within DMatrix
when inputs (like np.ndarray) don't already have column names. The procedure is removed to
avoid conflict with other inputs
[DMLC SGBoost #6605](https://github.com/dmlc/xgboost/pull/6605)

Our code is fully consistent with the open source XGBoost repo


awsbmillare:
looks good but let's not merge right away, we need to make sure we have the other branch stuff working in case we need to make changes/develop for current 1.3

awsbmillare (approved):

PR #264, opened 2022-04-07
Using ensemble flag for multi-model endpoint

dewan-c:
For multi-model inference we do not pass the ensemble flag when selecting provided model files. This leads to a mismatch between multi-model and single-model inference output.

This CR makes the change to use ensemble method for multi-model, unless explicitly turned off by the user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ src/sagemaker_xgboost_container.egg-info/
 __pycache__
 .coverage*
 .mypy_cache/
+.idea/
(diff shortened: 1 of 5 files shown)


awsbmillare:
what's the issue with the build failure?

dewan-c:
The first build failed because I left the base branch as master(1.5-1). I created a new PR and changed the base to 1.3-1 for the second one and it passed.

awsbmillare (approved):

PR #269, opened 2022-04-08
Add supported response accept encodings to algorithm_mode

awsbmillare:
*Issue #, if available:*

*Description of changes:*

Sagemaker xgboost inference server already supports responses in the format of application/json and text/csv when using script_mode. In algorithm_mode, however, the code is hard coded to always return in the format text/csv. This change simply extends this response format support to algorithm_mode. This was simply missed from previous developments.

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/serve.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/serve.py
@@ -22,6 +22,7 @@ from gunicorn.six import iteritems
 import flask
 import gunicorn.app.base
 
+from sagemaker_containers.beta.framework import encoders
 from sagemaker_xgboost_container.algorithm_mode import integration
 from sagemaker_xgboost_container.algorithm_mode import serve_utils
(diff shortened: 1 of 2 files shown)


src/sagemaker_xgboost_container/algorithm_mode/serve.py
```diff
@@ -225,19 +226,19 @@ def invocations():
...
+    if serve_utils.is_selectable_inference_output():
         return _handle_selectable_inference_response(preds, accept)
 
     if SAGEMAKER_BATCH:
         return_data = "\n".join(map(str, preds.tolist())) + '\n'
     else:
-        return_data = ",".join(map(str, preds.tolist()))
+        return_data = encoders.encode(preds.tolist(), accept)
 
     return flask.Response(response=return_data, status=http.client.OK, mimetype="text/csv")
```
NikhilRaverkar, line 243: Does this `mimetype` need to be updated too?

awsbmillare: good catch, updating on next commit

PR #273, opened 2022-04-12
Add warnings when validation files are suspected to be identical with training files

haixiw:
*Description of changes:*
Add warnings when validation files are suspected to be identical with training files

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -20,7 +20,8 @@ import xgboost as xgb
 from sklearn.model_selection import RepeatedKFold, RepeatedStratifiedKFold
 from sagemaker_algorithm_toolkit import exceptions as exc
 from sagemaker_algorithm_toolkit.channel_validation import Channel
-from sagemaker_xgboost_container.data_utils import get_content_type, get_dmatrix, get_size, validate_data_file_path
+from sagemaker_xgboost_container.data_utils import get_content_type, get_dmatrix, get_size, \
+    validate_data_file_path, check_data_redundancy
 from sagemaker_xgboost_container import distributed
 from sagemaker_xgboost_container import checkpointing
 from sagemaker_xgboost_container.algorithm_mode import channel_validation as cv
@@ -93,6 +94,8 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
         if train_path == validate_path or os.path.basename(train_path) == os.path.basename(validate_path):
             logger.warning('Found same path for training and validation. This is not recommended and results may not '
                            'be correct.')
+        else:
+            check_data_redundancy(train_path, validate_path)
         train_val_dmatrix = get_dmatrix([train_path, validate_path], content_type,
                                         csv_weights=csv_weights, is_pipe=is_pipe)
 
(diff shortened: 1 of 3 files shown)


src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -622,3 +622,39 @@ def _make_symlink(path, source_path, name, index):
...
+        if not os.path.exists(validate_path):
+            exc.UserError("validate path is not existed")
+
+    isRedundant = False
+    training_files_set = set(f for f in os.listdir(train_path) if os.path.isfile(os.path.join(train_path, f)))
+    validation_files_set = set(f for f in os.listdir(validate_path) if os.path.isfile(os.path.join(validate_path, f)))
+    same_name_files = training_files_set.intersection(validation_files_set)
+    if len(same_name_files) != 0:
+        for f in same_name_files:
+            f_train_name = os.path.join(train_path, f)
```
awsbmillare, line 649: nit: f_train_name is really f_train_path or like full_train_path, ditto for f_validate_name.

Also really f should be fname as its really a string (a name) and not a file object but I think that's fine here for concision.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -622,3 +622,39 @@ def _make_symlink(path, source_path, name, index):
...
+            f_validate_name = os.path.join(validate_path, f)
+            f_train_size = os.path.getsize(f_train_name)
+            f_validate_size = os.path.getsize(f_validate_name)
+            if f_train_size == f_validate_size:
+                isRedundant = True
+                logging.warning("Suspected identical files found. ({} and {} with same size {} bytes)."
+                                " Note: Duplicate data in the training set and validation set will"
```
awsbmillare, line 656: we can make a weaker claim here and just say "is usually not intentional and can impair validity of the model evaluation..."

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -93,6 +94,8 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
         if train_path == validate_path or os.path.basename(train_path) == os.path.basename(validate_path):
             logger.warning('Found same path for training and validation. This is not recommended and results may not '
                            'be correct.')
+        else:
```
awsbmillare, line 97: for readability maybe put a comment here? that this just warns cx for now, no conditional logic happens here

awsbmillare (changes requested):
Thanks for this work. It's great to help warn customers in this possibly common mistake case.

src/sagemaker_xgboost_container/algorithm_mode/train.py
haixiw: updated

test/unit/test_data_utils.py
```diff
@@ -244,3 +244,8 @@ def test_parse_protobuf_dmatrix_single_feature_label(self):
                 pb_path = os.path.join(self.data_path, 'recordio_protobuf', file_path)
                 reader = data_utils.get_recordio_protobuf_dmatrix
                 self._check_dmatrix(reader, pb_path, 1, 1)
+
```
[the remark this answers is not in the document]
haixiw: added a negative case

awsbmillare (approved):

awsbmillare:
why is there build failure? do we need to merge first to get to next stage so it builds correctly?

PR #277, opened 2022-04-19
Turn off OpenMP multi-threading for python processes

haixiw:
*Issue #, if available:*

*Description of changes:*
OpenMP does muti-threading within a process, the default number of processes is the number of CPU. 
We are using Python's multiprocessing module inside the container, So if there's 4 cores available and 4 processes initiated, we end up running 16 threads on 4 cores, which is not at peak efficiency.

Setting OMP_NUM_THREADS=1 basically turns off the OpenMP multi-threading, so each of your Python processes remains single-threaded.


src/sagemaker_xgboost_container/serving.py
```diff
@@ -43,6 +43,13 @@ def is_multi_model():
     return os.environ.get("SAGEMAKER_MULTI_MODEL")
 
 
+def set_default_env():
+    env_default_dict = {"OMP_NUM_THREADS": "1"}
```
mabunday, line 47: Please define `"1"` as a constant somewhere with a descriptive variable name.

src/sagemaker_xgboost_container/serving.py
```diff
@@ -43,6 +43,13 @@ def is_multi_model():
...
+    for key, value in env_default_dict.items():
+        if os.getenv(key) is None:
+            os.environ[key] = value
```
mabunday, line 50:
```suggestion
        user_specified_value = os.getenv(key)
        if user_specific_value is not None:
            os.environ[key] = user_specific_value
```

mabunday (changes requested):
Can you also add basic unit tests for these changes? 

Also, please rename the PR/commits to something more descriptive (unless this PR isn't complete). These changes only set defaults for `OMP_NUM_THREADS`, there's no multiprocessing optimization with these changes.

src/sagemaker_xgboost_container/serving.py
awsbmillare: its optimization of multiprocessing parameters since we are tuning threads per process given that process count normally equals CPU core count, but yes we should try to be more descriptive here of the motivation so that someone outside this conversation can see what's going on here. It needs to be standalone in understanding.

awsbmillare:
its optimization of multiprocessing parameters since we are tuning threads per process given that process count normally equals CPU core count, but yes we should try to be more descriptive here of the motivation so that someone outside this conversation can see what's going on here. It needs to be standalone in understanding.

mabunday:
Thanks for the context. How was 1 determined to be the optimal number? Adding details about experiments and tests would also be helpful.

awsbmillare:
Also, I don't think we will be able to do unit tests here, but we can do local integ tests. We need to make sure that when running through all code path variants of the inference server, that the OMP_NUM_THREADS variable is set to 1.

mabunday:
Can't those code paths be mocked? Local integration tests are fine too though if that's easier.

src/sagemaker_xgboost_container/serving.py
haixiw: Thanks for the reply but I am a little bit confused right here. So this function basically was to set ENV variables with default values if users haven't specified them.  

What do you mean by user_specific_value? (Correct me if I am wrong) From my understanding, OMP_NUM_THREADS = 1 should be a default setting for all inference tasks. That's why I use set_default_env for the function name. 

I will set "1" to a constant rather than hard coding

awsbmillare:
It's my opinion that once we get into the realm of cross language and cross process communication/signaling, its probably better to just verify on the ground truth as reconstructing the mocks for this doesn't as strongly guarantee the underlying logic actually fixes the problem.

Haixin, please call out the different inference server instantiation code paths. We need to be sure that all of them observe the desired env var change. If this is not easy to do in tests, we at least need to call out these paths and verify manually. If we don't call this out, we are continuing to hide the confusion we encountered when trying to figure out why certain messages weren't being logged.

awsbmillare (approved):

awsbmillare (approved):

awsbmillare:
Also comment in the PR that this env change only applies to serving/inference, not training. Having this set to training would be bad as we'd kill all parallelism/performance.

src/sagemaker_xgboost_container/serving.py
mabunday: @awsbmillare Perhaps a better function name if that's the intended behavior then is something like `set_unspecified_env_value_defaults`. 

In the original code, if the keys in `env_default_dict` aren't already defined in the environment then they'll be set to the default values. But if they are already defined (and many libraries set `OMP_NUM_THREADS`) then nothing happens - just want to check that this is intended and will work in all execution paths.

src/sagemaker_xgboost_container/constants/sm_env_constants.py
```diff
@@ -32,3 +32,7 @@
 SAGEMAKER_INFERENCE_OUTPUT = 'SAGEMAKER_INFERENCE_OUTPUT'
 SAGEMAKER_DEFAULT_INVOCATIONS_ACCEPT = 'SAGEMAKER_DEFAULT_INVOCATIONS_ACCEPT'
 SAGEMAKER_BATCH = 'SAGEMAKER_BATCH'
+
+# Multiprocessing related constants
+# Turn off OpenMP multi-threading, Let Python processes remain single-threaded
```
mabunday, line 37: This comment should be moved to line 48 in `serving.py` (and anywhere else we actually set the environment variable) instead of here.

`Single-thread processes by default. Multithreading can introduce significant performance overhead due to task switching.`

src/sagemaker_xgboost_container/serving.py
mabunday: Can users specify environment variables for the container? If so, does this change allow them to override and set a custom value for `OMP_NUM_THREADS`?

src/sagemaker_xgboost_container/constants/sm_env_constants.py
awsbmillare: Agreed, we want to make sure this is only changed in serve, not training

src/sagemaker_xgboost_container/serving.py
awsbmillare: I think its more we are grabbing the current env vars and putting it in the dict (to make it easy to walk through). If OMP_NUM_THREADs is undefined, then we want to set it to a default of 1. If it does have a setting, we take that as the customer wants to specify this and we let that be the value observed. This way the customer can still tune the thread count for optimal performance. 1 is likely not always optimal and depends on different factors.

test/unit/test_serving.py
```diff
@@ -102,6 +103,12 @@ def test_serving_entrypoint_start_gunicorn(mock_server):
     mock_server.start.assert_called_once()
 
 
+@patch('sagemaker_xgboost_container.serving.server')
+def test_serving_entrypoint_set_default_env(mock_server):
+    serving.serving_entrypoint()
+    assert os.getenv('OMP_NUM_THREADS') == sm_env_constants.ONE_THREAD_PER_PROCESS
+
+
```
mabunday, line 111: Thanks for the test. Let's add a "negative" tests for the other entrypoints that assert `os.getenv('OMP_NUM_THREADS') != sm_env_constants.ONE_THREAD_PER_PROCESS`

mabunday (changes requested):

src/sagemaker_xgboost_container/serving.py
```diff
@@ -43,6 +44,14 @@ def is_multi_model():
     return os.environ.get("SAGEMAKER_MULTI_MODEL")
 
 
+def set_default_env():
```
mabunday, line 47:
```suggestion
def set_default_serving_env_if_unspecified() -> None:
"""Set default values for environment variables if they aren't already specified."""
```

src/sagemaker_xgboost_container/serving.py
```diff
@@ -43,6 +44,14 @@ def is_multi_model():
...
+    env_default_dict = {"OMP_NUM_THREADS": sm_env_constants.ONE_THREAD_PER_PROCESS}
+    for key, value in env_default_dict.items():
+        user_specified_value = os.getenv(key)
+        if user_specified_value is None:
+            os.environ[key] = value
```
mabunday, line 52:
```suggestion
    for always_specified_key, default_value in env_default_dict.items():
        try:
            # If this does not throw, the user has specified a non-default value.
            os.environ[always_specified_key]
        except KeyError:
            #  Key that is always specified is not set in the environment. Set default value.
            os.environ[always_specified_key] = default_value
```

mabunday (changes requested):

test/unit/test_serving.py
```diff
@@ -102,6 +103,19 @@ def test_serving_entrypoint_start_gunicorn(mock_server):
...
+@patch('sagemaker_xgboost_container.serving.set_default_serving_env_if_unspecified')
+def test_serving_entrypoint_set_default_env(mock_set_default_serving_env_if_unspecified, mock_server):
+    serving.serving_entrypoint()
+    mock_set_default_serving_env_if_unspecified.assert_called_once()
+    assert os.getenv('OMP_NUM_THREADS') == sm_env_constants.ONE_THREAD_PER_PROCESS
+    with patch.dict(os.environ, {"OMP_NUM_THREADS": "USER_SPECIFIED_VALUE"}, clear=True):
+        mock_set_default_serving_env_if_unspecified.reset_mock()
+        serving.serving_entrypoint()
+        mock_set_default_serving_env_if_unspecified.assert_called_once()
+        assert os.getenv('OMP_NUM_THREADS') == "USER_SPECIFIED_VALUE"
```
mabunday, line 116: Please split into two tests, similar to this: 

```suggestion
def test_serving_entrypoint_set_default_env_positive(mock_set_default_serving_env_if_unspecified, mock_server):
    serving.serving_entrypoint()
    mock_set_default_serving_env_if_unspecified.assert_called_once()
    assert os.getenv('OMP_NUM_THREADS') == sm_env_constants.ONE_THREAD_PER_PROCESS
        
def test_serving_entrypoint_set_default_env_negative(mock_set_default_serving_env_if_unspecified, mock_server):
    with patch.dict(os.environ, {"OMP_NUM_THREADS": "USER_SPECIFIED_VALUE"}, clear=True):
        serving.serving_entrypoint()
        mock_set_default_serving_env_if_unspecified.assert_called_once()
        assert os.getenv('OMP_NUM_THREADS') == "USER_SPECIFIED_VALUE"
```

mabunday (changes requested):

mabunday:
Changes look good. Can you check the failing codebuild please?

mabunday (approved):

PR #280, opened 2022-04-19
Fix JSON output format

dewan-c:
`Description`

We recently enabled prediction response in JSON(`accept=application/json`), for algorithm mode, but the output doesn't match expected format in our documentation. This PR updates the output format.

existing output sample: `b'[0.43861907720565796, 0.4533972144126892, 0.06351257115602493]'`
new output sample: `b'{"predictions": [{"score": 0.43861907720565796}, {"score": 0.4533972144126892}, {"score": 0.06351257115602493}]}'`

Note that this doesn't include output format update for json format.

`Testing`

Unit tests
Local integ tests

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,13 @@ deps =
     xgboostlatest: xgboost
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
+conda_deps=
+    pyarrow==1.0.1
+    tbb==2020.2
+    mlio-py==0.7.0
+conda_channels=
+    conda-forge
+    mlio
 commands =
     pytest --cov=sagemaker_xgboost_container --cov-fail-under=60 test/unit # increase minimum bar over time (75%+)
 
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -16,6 +16,13 @@ deps =
     xgboostlatest: xgboost
     -r{toxinidir}/requirements.txt
     -r{toxinidir}/test-requirements.txt
+conda_deps=
```
awsbmillare, line 19: can you comment on why these were added?

awsbmillare (approved):

tox.ini
dewan-c: I was only able to run the tests successfully by specifying these dependency versions. Otherwise the versions installed by conda lead to errors during test run, for example `mlio` not being able to find `libtbb` file because the file had a different suffix for the version.
I wanted to see if others were able to run the unit tests locally, without any issues. I can drop this out of the CR if noone else had any issues.

awsbmillare: got it, I think its fine, just makes things more robust to building on different setups.

PR #285, opened 2022-04-27
Case2272044573 master merge

NikhilRaverkar:
*Issue #, if available:*
* Fix for a FileExistsError error when both train and validation dataset contain files with the same name
*Description of changes:*
* Avoiding duplicate file collision
* Add warnings when validation files are suspected to be identical (by @haixiw ) Ref: [PR](https://github.com/aws/sagemaker-xgboost-container/pull/273) 


src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +616,48 @@ def get_files_path(data_path):
...
+    if not os.path.exists(train_path):
+        exc.UserError("training data's path is not existed")
+    if not os.path.exists(validate_path):
+        exc.UserError("validation data's path is not existed")
+
+    isRedundant = False
+    training_files_set = set(f for f in os.listdir(train_path) if os.path.isfile(os.path.join(train_path, f)))
+    validation_files_set = set(f for f in os.listdir(validate_path) if os.path.isfile(os.path.join(validate_path, f)))
+    same_name_files = training_files_set.intersection(validation_files_set)
+    if same_name_files:
```
mabunday, line 650: This if statement can be removed.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +616,48 @@ def get_files_path(data_path):
...
     return files_path
+
+
+def _make_symlink(path, source_path, name, index):
+    base_name = os.path.join(source_path, name + '_' + str(index))
+    logging.info('creating symlink between Path {} and destination {}'.format(source_path, base_name))
+    os.symlink(path, base_name)
+
+
+def check_data_redundancy(train_path, validate_path):
```
mabunday, line 627: Is the return value used for anything aside from the unit tests? If not, let's check instead that logging.warning is called in place of the true check.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +616,48 @@ def get_files_path(data_path):
...
+    """Give a hint about whether same files exist in the train/validation folder.
```
mabunday, line 628:
```suggestion
    """Log a warning if suspected duplicate files are found in the training and validation folders.
```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +616,48 @@ def get_files_path(data_path):
...
+
+    The validation score of models would make less sense if there are same data in both training/validation folder.
```
mabunday, line 630:
```suggestion
    The validation score of models would be invalid if the same data is used for both training and validation.
```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +616,48 @@ def get_files_path(data_path):
...
+    This function is used to give warnings when there are files sharing the same name and file size in training and
+    validation folders
```
mabunday, line 632:
```suggestion
    Files are suspected of being duplicates when the file names are the same and their sizes are the same.
```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +616,48 @@ def get_files_path(data_path):
                 break
 
     return files_path
+
+
+def _make_symlink(path, source_path, name, index):
+    base_name = os.path.join(source_path, name + '_' + str(index))
+    logging.info('creating symlink between Path {} and destination {}'.format(source_path, base_name))
```
mabunday, line 623: If the minimum Python version on the container is Python 3.6, let's use f-strings here instead: https://realpython.com/python-f-strings/ Same in other places.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +616,48 @@ def get_files_path(data_path):
                 break
 
     return files_path
+
+
+def _make_symlink(path, source_path, name, index):
+    base_name = os.path.join(source_path, name + '_' + str(index))
```
mabunday, line 622:
```suggestion
    base_name = os.path.join(source_path, f"{name}_{str(index)}")
```

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -92,6 +93,12 @@ def get_validated_dmatrices(train_path, validate_path, content_type, csv_weights
     train_val_dmatrix = train_dmatrix
     if combine_train_val and train_dmatrix is not None and val_dmatrix is not None:
         logging.info("Read both train and validation data into one DMatrix")
+        if train_path == validate_path or os.path.basename(train_path) == os.path.basename(validate_path):
+            logger.warning('Found same path for training and validation. This is not recommended and results may not '
+                           'be correct.')
+        else:
+            # Check if there is potential data redundancy between training and validation sets
+            check_data_redundancy(train_path, validate_path)
```
mabunday, line 101: Seems like it would make more sense to make these checks and throw these warnings right after the construction of `train_path` and `validate_path`: https://github.com/aws/sagemaker-xgboost-container/blob/b42040bf32cb8b59c444c0f92c3bef44c29ca902/src/sagemaker_xgboost_container/training.py#L56-L57

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +485,52 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
         raise exc.UserError("Failed to load recordio-protobuf data with exception:\n{}".format(e))
 
 
+def _get_pipe_mode_files_path(data_path):
```
mabunday, line 488:
```suggestion
def _get_pipe_mode_files_path(data_path: Union[List[str], str]) -> List[str]:
```

Same below. Let's incrementally make our functions less ambiguous with type hints.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +485,52 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
...
+    """
+     :param data_path: Either directory or file
+    """
```
mabunday, line 491:
```suggestion
    """
    :param data_path: Either directory or file
    """
```

Same below.

mabunday (changes requested):

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: @haixiw can you please take a look? Thanks

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: This statement is setting `isRedundant = True`. Without an if statement, we can't return True/False from this function.

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: Changing.

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: changing in the next commit

src/sagemaker_xgboost_container/algorithm_mode/train.py
NikhilRaverkar: @haixiw can you please change this too?

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: updated

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: updated at one other place that was concatenating a string

src/sagemaker_xgboost_container/data_utils.py
haixiw: updated

test/unit/test_data_utils.py
```diff
@@ -225,3 +226,16 @@ def test_parse_protobuf_dmatrix_single_feature_label(self):
                 pb_path = os.path.join(self.data_path, 'recordio_protobuf', file_path)
                 reader = data_utils.get_recordio_protobuf_dmatrix
                 self._check_dmatrix(reader, pb_path, 1, 1)
+
+    @patch("logging.warning")
+    def test_check_data_redundancy(self, mock_log_warning):
```
mabunday, line 231: Please split this into two unit tests instead so that each test executes one code path.

src/sagemaker_xgboost_container/data_utils.py
mabunday: Looks like this still needs to be updated here.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +487,52 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
...
+        if not os.path.exists(f"${data_path}_0"):
+            logging.info('Pipe path {} does not exist!'.format(data_path))
+            return None
+    return files_path
+
+
+def _get_file_mode_files_path(data_path: Union[List[str], str]) -> List[str]:
+    """
+     :param data_path: Either directory or file
+    """
```
mabunday, line 508:
```suggestion
    """
    :param data_path: Either directory or file
    """
```

src/sagemaker_xgboost_container/data_utils.py
mabunday: Got it. Looks like it can be removed now though.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +618,40 @@ def get_files_path(data_path):
...
+            f_train_path = os.path.join(train_path, f)
+            f_validate_path = os.path.join(validate_path, f)
+            f_train_size = os.path.getsize(f_train_path)
+            f_validate_size = os.path.getsize(f_validate_path)
+            if f_train_size == f_validate_size:
+                logging.warning("Suspected identical files found. ({} and {} with same size {} bytes)."
+                                " Note: Duplicate data in the training set and validation set is usually"
+                                " not intentional and can impair the validity of the model evaluation by"
+                                " the validation score."
+                                .format(f_train_path, f_validate_path, f_validate_size))
```
mabunday, line 657: Please use f-strings here as well.

mabunday (changes requested):

mabunday:
There's been a lot of commits on this - let's convert it to a draft until its ready for another round of reviews to avoid creating noise.

src/sagemaker_xgboost_container/training.py
```diff
@@ -54,7 +55,16 @@ def run_algorithm_mode():
         checkpoint_config = {}
 
     train_path = os.environ[sm_env_constants.SM_CHANNEL_TRAIN]
-    val_path = os.environ.get(sm_env_constants.SM_CHANNEL_VALIDATION)
+    val_path = os.environ.get(sm_env_constants.SM_CHANNEL_VALIDATION,
+                              sm_env_constants.SM_CHANNEL_VALIDATION_NOT_SET)
+
+    if val_path != sm_env_constants.SM_CHANNEL_VALIDATION_NOT_SET:
```
mabunday, line 61: What's the advantage of using `SM_CHANNEL_VALIDATION_NOT_SET` over just checking for `None`? If `SM_CHANNEL_VALIDATION` does not exist in the environment, then `val_path` will be set to `None` in which case the check on 61 can just be `if val_path is not None:`.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +618,39 @@ def get_files_path(data_path):
...
+    The validation score of models would be invalid if the same data is used for both training and validation.
+    Files are suspected of being duplicates when the file names are the same and their sizes are the same.
+
+    param train_path : path to training data
+    param validate_path : path to validation data
+    """
+    if not os.path.exists(train_path):
+        exc.UserError("training data's path is not existed")
+    if not os.path.exists(validate_path):
+        exc.UserError("validation data's path is not existed")
```
mabunday, line 641:
```suggestion
    if not os.path.exists(train_path):
        exc.UserError(f"Training data path '{train_path}' does not exist. Please ensure the environment variable '{sm_env_constants.SM_CHANNEL_TRAIN}' is set appropriately.")
    if not os.path.exists(validate_path):
        exc.UserError("Validation data path '{validate_path}' does not exist. Please ensure the environment variable '{sm_env_constants.SM_CHANNEL_VALIDATION}' is set appropriately.")
```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -595,3 +618,39 @@ def get_files_path(data_path):
                 break
 
     return files_path
+
+
+def _make_symlink(path, source_path, name, index):
+    base_name = os.path.join(source_path, f"{name}_{str(index)}")
+    logging.info(f'creating symlink between Path {source_path} and destination {base_name}')
```
mabunday, line 625:
```suggestion
    logging.info(f'creating symlink between path {source_path} and destination {base_name}')
```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +487,52 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
...
+    """
+    :param data_path: Either directory or file
+    """
+    # For pipe mode, we leverages mlio directly by creating a list of SageMakerPipe.
+    if isinstance(data_path, list):
+        files_path = data_path
+    else:
+        files_path = [data_path]
+        if not os.path.exists(f"${data_path}_0"):
+            logging.info('Pipe path {} does not exist!'.format(data_path))
```
mabunday, line 500:
```suggestion
            logging.info(f"Pipe path {data_path} does not exist!")
```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +487,52 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
...
+
+def _get_file_mode_files_path(data_path: Union[List[str], str]) -> List[str]:
+    """
+    :param data_path: Either directory or file
+    """
+    # In file mode, we create a temp directory with symlink to all input files or
+    # directories to meet XGB's assumption that all files are in the same directory.
+
+    if isinstance(data_path, list):
+        logging.info('File path {} of input files'.format(data_path))
```
mabunday, line 513:
```suggestion
        logging.info("File path {data_path} of input files")
```

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -485,6 +487,52 @@ def get_recordio_protobuf_dmatrix(path, is_pipe=False):
...
+                return None
+            if os.path.isfile(path):
+                _make_symlink(path, files_path, os.path.basename(path), index)
+            else:
+                for file in os.scandir(path):
+                    _make_symlink(file, files_path, file.name, index)
+
+    else:
+        if not os.path.exists(data_path):
+            logging.info('File path {} does not exist!'.format(data_path))
```
mabunday, line 529:
```suggestion
            logging.info("File path {data_path} does not exist!")
```

src/sagemaker_xgboost_container/data_utils.py
mabunday: Is this for debugging? Not sure this log provides a lot of value here so consider removing it or updating the message to be a little more descriptive.

mabunday (changes requested):

src/sagemaker_xgboost_container/data_utils.py
NikhilRaverkar: This can be removed now that we have warning and validation logic in place

PR #300, opened 2022-07-14
Reverting numpy 1.21.2 to 1.19.2 and adding regression

NikhilRaverkar:
Issue #, if available:
* fixing https://github.com/aws/sagemaker-scikit-learn-container/issues/106 by reverting numpy version.
*  Added regression to verify the fix and catch the re occurrence of the issue. 
* Although the issue was reported in scikitlearn container, XGBoost uses same version of PyArrow. SO this is just a precautionary change. 

Description of changes:
* Numpy 1.2.x broke the integration with Pyarrow 1.0. Rolling back numpy version to 1.19.2
* Added regression test

### DIFF
--- a/test/unit/test_data_utils.py
+++ b/test/unit/test_data_utils.py
@@ -14,6 +14,7 @@ from __future__ import absolute_import
 import unittest
 import os
 from pathlib import Path
+import pandas as pd
 import shutil
 import signal
 import subprocess
@@ -252,3 +253,7 @@ class TestTrainUtils(unittest.TestCase):
         pb_file_paths = ['pb_files']
         with self.assertRaises(Exception):
             data_utils.check_data_redundancy(pb_file_paths[0], pb_file_paths[1])
+
+    def test_pyarrow_to_parquet_conversion_does_not_throw_exception(self):
+        df = pd.DataFrame({'x': [1, 2]})
+        df.to_parquet('test.parquet', engine='pyarrow')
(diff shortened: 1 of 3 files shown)


test/unit/test_data_utils.py
```diff
@@ -252,3 +253,12 @@ def test_check_data_redundancy_throws_exception_pipe(self):
         pb_file_paths = ['pb_files']
         with self.assertRaises(Exception):
             data_utils.check_data_redundancy(pb_file_paths[0], pb_file_paths[1])
+            
+    def test_pyarrow_to_parquet_conversion_does_not_throw_exception(self):
+        try:
+             df = pd.DataFrame({'x': [1,2]})
+             df.to_parquet('test.parquet', engine='pyarrow')
+        except Exception as e:
+            assert False, "conversion to parquet should not fail"
```
mabunday, line 262:
```suggestion
    def test_pyarrow_to_parquet_conversion_regression_issue_106(self):
        df = pd.DataFrame({'x': [1,2]})
        df.to_parquet('test.parquet', engine='pyarrow')
```

We don't want the exception to be swallowed if it does fail.

mabunday: If you want to add an assertion, you check check that the values are equivalent before and after conversion.

NikhilRaverkar: On the exception it will fail the test, thus will fail the build as well.

mabunday: Yes, but surfacing the actual exception will make the failure obvious and debugging easier instead of swallowing it with a generic "assert False".

mabunday (approved):

PR #72, opened 2019-11-26
Add debuggging

edwardjkim:
*Issue #, if available:*

*Description of changes:*

I don't want to reveal too much information here. Please let me know if you would like to see docs and notebooks.

Tests: flake8, pytest test/unit, internal container tests, internal integration tests, local containers in #73 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #301, opened 2022-08-25
Mix-in smdebug.xgboost.Hook and xgb.callback.TrainingCallback for callback compatibility in XGB-3.0.0+

haixiw:
*Issue #, if available:*

*Description of changes:*
from xgboost version 1.3.0 and later, all the callbacks consumed by xgb.train() must be instances of xgb.callback.TrainingCallback

we can't mixed new callbacks with Hook callback from smdebugger

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -273,7 +273,7 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                     early_stopping_metric=early_stopping_metric, early_stopping_rounds=early_stopping_rounds,
                     save_model_on_termination=save_model_on_termination, is_master=is_master, fold=len(bst))
                 callbacks.append(SmeDebugHook(hyperparameters=train_cfg, train_dmatrix=train_dmatrix,
-                                          val_dmatrix=val_dmatrix))
+                                              val_dmatrix=val_dmatrix))
                 evals_result = {}
                 logging.info("Train cross validation fold {}".format((len(bst) % kfold) + 1))
                 booster = xgb.train(train_cfg, cv_train_dmatrix, num_boost_round=num_round-iteration,
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -24,6 +24,7 @@ from sagemaker_xgboost_container.data_utils import get_content_type, get_dmatrix
  check_data_redundancy
 from sagemaker_xgboost_container import distributed
 from sagemaker_xgboost_container import checkpointing
+from sagemaker_xgboost_container.callback import SmeDebugHook
 from sagemaker_xgboost_container.algorithm_mode import channel_validation as cv
 from sagemaker_xgboost_container.algorithm_mode import hyperparameter_validation as hpv
 from sagemaker_xgboost_container.algorithm_mode import metrics as metrics_mod
@@ -231,7 +232,8 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                 checkpoint_dir=checkpoint_dir, early_stopping_data_name=early_stopping_data_name,
                 early_stopping_metric=early_stopping_metric, early_stopping_rounds=early_stopping_rounds,
                 save_model_on_termination=save_model_on_termination, is_master=is_master)
-
+            callbacks.append(SmeDebugHook(hyperparameters=train_cfg, train_dmatrix=train_dmatrix,
+                                          val_dmatrix=val_dmatrix))
             bst = xgb.train(train_cfg, train_dmatrix, num_boost_round=num_round-iteration, evals=watchlist,
                             feval=configured_feval, callbacks=callbacks, xgb_model=xgb_model, verbose_eval=False)
 
@@ -270,7 +272,8 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                     checkpoint_dir=checkpoint_dir, early_stopping_data_name=early_stopping_data_name,
                     early_stopping_metric=early_stopping_metric, early_stopping_rounds=early_stopping_rounds,
                     save_model_on_termination=save_model_on_termination, is_master=is_master, fold=len(bst))
-
+                callbacks.append(SmeDebugHook(hyperparameters=train_cfg, train_dmatrix=train_dmatrix,
+                                          val_dmatrix=val_dmatrix))
                 evals_result = {}
                 logging.info("Train cross validation fold {}".format((len(bst) % kfold) + 1))
                 booster = xgb.train(train_cfg, cv_train_dmatrix, num_boost_round=num_round-iteration,
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -32,6 +32,7 @@ from sagemaker_xgboost_container.callback import add_debugging
 from sagemaker_xgboost_container.constants.xgb_constants import CUSTOMER_ERRORS, XGB_MAXIMIZE_METRICS
 from sagemaker_xgboost_container.constants.sm_env_constants import SM_OUTPUT_DATA_DIR
 from sagemaker_xgboost_container.prediction_utils import ValidationPredictionRecorder
+from sagemaker_xgboost_container.constants.xgb_constants import XGB_VERSION
 
 MODEL_NAME = "xgboost-model"
 
@@ -232,9 +233,13 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                 checkpoint_dir=checkpoint_dir, early_stopping_data_name=early_stopping_data_name,
                 early_stopping_metric=early_stopping_metric, early_stopping_rounds=early_stopping_rounds,
                 save_model_on_termination=save_model_on_termination, is_master=is_master)
-            add_debugging(callbacks=callbacks, hyperparameters=train_cfg, train_dmatrix=train_dmatrix,
-                          val_dmatrix=val_dmatrix)
-
+            if XGB_VERSION < "1.3.0":
+                # when xgboost >= 1.3.0, it doesn't allow callback which are not instances 
+                # of callback.TrainingCallback
+                # ref : https://github.com/dmlc/xgboost/blob/v1.5.2/python-package/xgboost/training.py#L62
+                #       https://xgboost.readthedocs.io/en/latest/python/callbacks.html
+                add_debugging(callbacks=callbacks, hyperparameters=train_cfg, train_dmatrix=train_dmatrix,
+                            val_dmatrix=val_dmatrix)
             bst = xgb.train(train_cfg, train_dmatrix, num_boost_round=num_round-iteration, evals=watchlist,
                             feval=configured_feval, callbacks=callbacks, xgb_model=xgb_model, verbose_eval=False)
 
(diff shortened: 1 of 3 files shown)


src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -273,9 +278,9 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
                     checkpoint_dir=checkpoint_dir, early_stopping_data_name=early_stopping_data_name,
                     early_stopping_metric=early_stopping_metric, early_stopping_rounds=early_stopping_rounds,
                     save_model_on_termination=save_model_on_termination, is_master=is_master, fold=len(bst))
-                add_debugging(callbacks=callbacks, hyperparameters=train_cfg, train_dmatrix=cv_train_dmatrix,
-                              val_dmatrix=cv_val_dmatrix)
-
+                if XGB_VERSION < "1.3.0":
```
mabunday, line 281: `XGB_VERSION` is a hardcoded constant so this condition will never be satisfied, right? Should we just remove the call to `add_debugging` entirely instead?

src/sagemaker_xgboost_container/constants/xgb_constants.py
```diff
@@ -11,6 +11,8 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
+XGB_VERSION = "1.5.2"
```
mabunday, line 14: Instead of hardcoding a version, consider using something like this https://stackoverflow.com/a/32965521

src/sagemaker_xgboost_container/algorithm_mode/train.py
haixiw: I was thinking about it,but if sagemaker debugger supports xgboost>1.3.0 in the future, this will cost little to update

src/sagemaker_xgboost_container/constants/xgb_constants.py
haixiw: will update

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -232,9 +233,13 @@ def train_job(train_cfg, train_dmatrix, val_dmatrix, train_val_dmatrix, model_di
...
                 early_stopping_metric=early_stopping_metric, early_stopping_rounds=early_stopping_rounds,
                 save_model_on_termination=save_model_on_termination, is_master=is_master)
-            add_debugging(callbacks=callbacks, hyperparameters=train_cfg, train_dmatrix=train_dmatrix,
-                          val_dmatrix=val_dmatrix)
-
+            if XGB_VERSION < "1.3.0":
+                # when xgboost >= 1.3.0, it doesn't allow callback which are not instances 
+                # of callback.TrainingCallback
+                # ref : https://github.com/dmlc/xgboost/blob/v1.5.2/python-package/xgboost/training.py#L62
+                #       https://xgboost.readthedocs.io/en/latest/python/callbacks.html
```
mabunday, line 240: Per the code and doc links, can we solve this issue just by wrapping the offending callbacks as TrainingCallbacks? https://xgboost.readthedocs.io/en/latest/python/callbacks.html#defining-your-own-callback

That way we wouldn't need this check at all.

haixiw: it's from sagemaker debugger. https://github.com/awslabs/sagemaker-debugger/blob/master/smdebug/xgboost/hook.py#L36

haixiw: oh, wait , you mean wrap the hook by TrainingCallbacks? it should work but I think this sounds more like the work for debugger team. they only supports xgboost 0.90-2, 1.0-1 from readme. If we really need it, we can implemement our own callback instead.

mabunday: Yeah, exactly. I agree that ideally this should be a change made in smedebug, but I think it's worth evaluating the level of effort to convert the old API to the new one. This is the new API spec: https://xgboost.readthedocs.io/en/stable/python/python_api.html#callback-api

We could define a module similar to this: https://github.com/aws/sagemaker-xgboost-container/tree/b42040bf32cb8b59c444c0f92c3bef44c29ca902/src/sagemaker_xgboost_container/dmlc_patch that defines `class SmeDebugHook(xgb.callback.TrainingCallback)`. I'm not even sure we need to override the methods: https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/callback.py#L36 in order to get it to work, but that should be tested.

Let's open an issue with the SageMaker Debugger team https://github.com/awslabs/sagemaker-debugger/issues so they're aware of this issue, in any case.

dewan-c:
The debugger is what produces the profiling report right? Added Malav, as he's been working on ensuring profiling report works for v1.3-1 and up.

dewan-c:
Malav - did you use algorithm mode when verifying the reports are being generated?

malav-shastri:
> Malav - did you use algorithm mode when verifying the reports are being generated?

Hey Dewan - just checked, I have used Open Source distributed script mode(Framework mode) when verifying.

src/sagemaker_xgboost_container/callback.py
```diff
@@ -1,29 +1,27 @@
...
 
 logger = logging.getLogger(__name__)
 
-
-def add_debugging(callbacks, hyperparameters, train_dmatrix,
-                  val_dmatrix=None, json_config_path=None):
-    """Add a sagemaker debug hook to a list of callbacks.
-    :param callbacks: List of callback functions.
+class SmeDebugHook(xgb.callback.TrainingCallback, Hook):
+    """wrap sagemaker debug hook to TrainingCallback style
```
mabunday, line 9: Thanks for the change! Interested to see if this works.

```suggestion
    """Mix-in callback class. smedebug.xgboost.Hook uses legacy callback style and since XGB-3.0.0 mixing legacy callback instances with new TrainingCallback instances is not allowed. 
    
    See: https://github.com/dmlc/xgboost/blob/v1.3.0/python-package/xgboost/training.py#L92-L93
```

src/sagemaker_xgboost_container/callback.py
```diff
@@ -1,29 +1,32 @@
...
+        try:
+            Hook.__init__(self)
+            self.hyperparameters = hyperparameters
+            self.train_data = train_dmatrix
+            if val_dmatrix is not None:
+                self.validation_data = val_dmatrix
+            logging.info("Debug hook initialized")
+        except Exception as e:
+            logging.debug("Failed to create debug hook", e)
+            return
```
mabunday, line 32: Sorry, I should have caught this initially, but this implementation is actually a bit different from the existing because 

1. we no longer take the `json_config_path` to create the hook 
2. the initialized `SmeDebugHook` will always be appended to `callbacks`, even if an exception is thrown in `__init__` 

Here's what I just tested in ipython:

```python3
class A:
    pass

class B:
    def some_function(self, foo):
        return foo

class MixIn(A, B):
    pass

# MixIn instance has access to inherited function even without calling its parent init. Returns "some input"
MixIn().some_function("some input")
isinstance(MixIn(), A)  # True
isinstance(MixIn(), B)  # Also true
```

Therefore, can you try this instead?

```python3
class SmeDebugHook(xgb.callback.TrainingCallback, Hook):
    """Mix-in callback class. smedebug.xgboost.Hook uses legacy callback style
    and since XGB-3.0.0 mixing legacy callback instances with new TrainingCallback
    instances is not allowed.
    
    See: https://github.com/dmlc/xgboost/blob/v1.3.0/python-package/xgboost/training.py#L92-L93
    """"
    pass

def add_debugging(callbacks, hyperparameters, train_dmatrix,
                  val_dmatrix=None, json_config_path=None):
    """Add a sagemaker debug hook to a list of callbacks.
    :param callbacks: List of callback functions.
    :param hyperparameters: Dict of hyperparamters.
                            Same as `params` in xgb.train(params, dtrain).
    :param train_dmatrix: Training data set.
    :param val_dmatrix: Validation data set.
    :param json_config_path: If specified, this json config will be used
                             instead of default config file.
    """
    try:
        hook = SmeDebugHook.hook_from_config(json_config_path)
        hook.hyperparameters = hyperparameters
        hook.train_data = train_dmatrix
        if val_dmatrix is not None:
            hook.validation_data = val_dmatrix
        logging.info("Debug hook created from config")
    except Exception as e:
        logging.debug("Failed to create debug hook", e)
    else:
        callbacks.append(hook)
```

And then you can just revert the changes in `train.py`, too. 

I think you could also write this if you prefer:

```python3
class SmeDebugHook(xgb.callback.TrainingCallback, Hook):
    """Mix-in callback class. smedebug.xgboost.Hook uses legacy callback style
    and since XGB-3.0.0 mixing legacy callback instances with new TrainingCallback
    instances is not allowed.
    
    See: https://github.com/dmlc/xgboost/blob/v1.3.0/python-package/xgboost/training.py#L92-L93

    :param json_config_path: If specified, this json config will be used
                             instead of default config file.
    :param hyperparameters: Dict of hyperparamters.
                            Same as `params` in xgb.train(params, dtrain).
    :param train_dmatrix: Training data set.
    :param val_dmatrix: Validation data set.
    """"
    def __init__(self, json_config_path, hyperparameters, train_dmatrix, val_dmatrix):
        self = self.hook_from_config(json_config_path)
        self.hyperparameters = hyperparameters
        self.train_data = train_dmatrix
        if val_dmatrix is not None:
            self.validation_data = val_dmatrix

def add_debugging(callbacks, hyperparameters, train_dmatrix,
                  val_dmatrix=None, json_config_path=None):
    """Add a sagemaker debug hook to a list of callbacks.
    :param callbacks: List of callback functions.
    :param hyperparameters: Dict of hyperparamters.
                            Same as `params` in xgb.train(params, dtrain).
    :param train_dmatrix: Training data set.
    :param val_dmatrix: Validation data set.
    :param json_config_path: If specified, this json config will be used
                             instead of default config file.
    """
    try:
        hook = SmeDebugHook(json_config_path, hyperparameters, train_dmatrix, val_dmatrix)
        logging.info("Debug hook created from config")
    except Exception as e:
        logging.debug("Failed to create debug hook", e)
    else:
        callbacks.append(hook)
```

It's the call to `hook_from_config` that will throw if `json_config_path` does not exist: https://github.com/awslabs/sagemaker-debugger/blob/4a470d9c7711e6ff89e5557d91260ee86a0b71b3/smdebug/core/json_config.py#L80 so we still want to make sure we avoid appending the callback if this happens.

mabunday (changes requested):

src/sagemaker_xgboost_container/callback.py
haixiw: Oh Cool, Thanks!. Sorry I should've been more careful about that try ... catch .. . I was thinking as we never specify json_path when calling, so try to initialize in a clearer way. I will update in next commit
```
            add_debugging(callbacks=callbacks, hyperparameters=train_cfg, train_dmatrix=train_dmatrix,
                          val_dmatrix=val_dmatrix)
```

mabunday: Oh, that's a good point. 

Can you just do a quick search and see if we ever call `add_debugging` with `json_config_path`? It may be in some notebooks, perhaps, or example scripts. If you can't find any references, then let's initialize using your initial approach since I saw this comment here: https://github.com/awslabs/sagemaker-debugger/blob/4a470d9c7711e6ff89e5557d91260ee86a0b71b3/smdebug/xgboost/hook.py#L156

So we can still have the `add_debugging` function, but in `SmeDebugHook` instead of calling `self = self.hook_from_config(json_config_path)` we would do what you did originally with `Hook.__init__(self)`.

haixiw: I think the reason why it didn't throw FileNotFound Error is becasue there's default json file existing 
https://github.com/awslabs/sagemaker-debugger/blob/4a470d9c7711e6ff89e5557d91260ee86a0b71b3/smdebug/core/json_config.py#L88

> path = os.getenv(CONFIG_FILE_PATH_ENV_STR, DEFAULT_CONFIG_FILE_PATH)

mabunday: This is the original PR: https://github.com/aws/sagemaker-xgboost-container/pull/72 I'll look internally and see if I can find anything pertinent.

mabunday (approved):
Looks good! Thank you for the quick changes!

As a final step, please update the final commit message to something like "Mix-in smdebug.xgboost.Hook and xgb.callback.TrainingCallback for callback compatibility in XGB-3.0.0+"

I'll defer the exact wording to you - feel free to adjust accordingly. :)

PR #306, opened 2022-09-07
Fix custom metric miscalculation

haixiw:
*Description of changes:*
We need to maintain the order of **`custom_metric_list`** to prevent miscalculation of custom metrics in distributed training scenario. 

### DIFF
--- a/src/sagemaker_xgboost_container/metrics/custom_metrics.py
+++ b/src/sagemaker_xgboost_container/metrics/custom_metrics.py
@@ -239,7 +239,11 @@ CUSTOM_METRICS = {
 
 
 def get_custom_metrics(eval_metrics):
-    """Get container defined metrics from metrics list."""
+    """Get container defined metrics from metrics list.
+
+    The order of the returning custom metrics need to be consistent with the input for distributed training.
+    Otherwise, metrics reported from each host will be miscalculated in the master host. (P70679777)
+    """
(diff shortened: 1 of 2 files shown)


mabunday (changes requested):
Thanks for the change, Haixin. Let's go one step further and make `CUSTOM_METRICS` an `OrderedDict` object instead of a regular dictionary object: https://github.com/aws/sagemaker-xgboost-container/blob/4723e36c43ae20fa4decacd1f1589a218eac70fc/src/sagemaker_xgboost_container/metrics/custom_metrics.py#L222

Please also add one or two simple unit tests for this function that check:

1. Metrics not in `CUSTOM_METRICS.keys()` are not added to the return list
2. Metrics are returned in the correct order. This can be used: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual

Sets are also used here and here (please check for other usages as well):
1. https://github.com/aws/sagemaker-xgboost-container/blob/196bc874127101951365dc5d39e10bf3db494c61/src/sagemaker_xgboost_container/algorithm_mode/train_utils.py
2. https://github.com/aws/sagemaker-xgboost-container/blob/e4912187a87c546ec0e1f9d36031eca0f92a78f1/src/sagemaker_xgboost_container/algorithm_mode/serve_utils.py

Do these need to be updated as well?

mabunday (approved):

PR #320, opened 2022-09-12
Maintain the order of union metric list in distributed training

haixiw:
*Issue #, if available:*

*Description of changes:*
A followup for #306 , need to maintain the order of union metrics if customer specifies eval_metric and tuning metric

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/train_utils.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train_utils.py
@@ -33,7 +33,9 @@ def get_union_metrics(metric_a, metric_b):
     elif metric_b is None:
         return metric_a
     else:
-        metric_list = list(set(metric_a).union(metric_b))
+        # The order of metric_list need to be consistent among all hosts in distributed training
+        # So we have metric_list sorted here.
+        metric_list = sorted(list(set(metric_a).union(metric_b)))
         return metric_list
 
 


src/sagemaker_xgboost_container/algorithm_mode/train_utils.py
```diff
@@ -33,7 +33,9 @@ def get_union_metrics(metric_a, metric_b):
     elif metric_b is None:
         return metric_a
     else:
-        metric_list = list(set(metric_a).union(metric_b))
+        # The order of metric_list need to be consistent among all hosts in distributed training
+        # So we have metric_list sorted here.
+        metric_list = sorted(list(set(metric_a).union(metric_b)))
```
mabunday, line 38: Not sure if it matters, but this is returning a generator, which means it can be exhausted unlike a list. I would recommend one of the following two changes:

```suggestion
        sorted_union_metric_list = list(sorted(list(set(metric_a).union(metric_b))))
        return sorted_union_metric_list
```

or

```suggestion
        union_metric_list = metric_a + [metric for metric in metric_b if metric not in metric_a]
        return union_metric_list
```

mabunday (changes requested):

src/sagemaker_xgboost_container/algorithm_mode/train_utils.py
haixiw: Hi, Thanks for the comment, by this doc : https://docs.python.org/3/library/functions.html#sorted sorted() return a list. And I also try to print(type(sorted(metric_list))) and got a list type. 
And for union_metric_list = metric_a + [metric for metric in metric_b if metric not in metric_a] I think set's union method has a better time complexity as they are using hash values.

mabunday: Ah, you're right - my bad! LGTM then.

mabunday (approved):

PR #342, opened 2022-11-08
Installed required dependency and fix some package versions

dewan-c:
Last Codebuild for 1.0-1 was failing due to missing dependency(glog), and incompatible dependency version(tbb). Explicitly installing these from conda-forge, as they don't seem to be installed automatically. 
Also updated some package versions which had conflicts during build. 
Added buildspec. The codebuild project is configured to only run when PR is merged.

*Testing:*
* Unit tests with `tox` passed.
* Local integ tests passed.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -131,7 +131,7 @@ dependencies.
 
     git clone https://github.com/aws/sagemaker-xgboost-container.git
     cd sagemaker-xgboost-container
-    pip install -e .[test]
+    pip install -e .[test] # Use bash when running this command
 
 Conda is also required and can be installed by following the instructions at https://conda.io/projects/conda/en/latest/user-guide/install/index.html. For convenience, the Linux installation commands are provided as an example.
 
--- a/ci/buildspec.yml
+++ b/ci/buildspec.yml
@@ -1,0 +1,61 @@
+version: 0.2
+
+phases:
+  install:
+    runtime-versions:
+      python: 3.6
+      docker: 19
+  pre_build:
+    commands:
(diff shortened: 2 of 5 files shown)


mabunday:
Just as an additional sanity check, can you run a training integration test with parquet? Not sure if the local integration tests catch the integration issues between Protobuf, ML-IO, and PyArrow or not.

amzn-choeric:
Was this branch ever properly integrated with CodeBuild? As in, is there still a follow-up change required to make the builds successful?

dewan-c:
> Just as an additional sanity check, can you run a training integration test with parquet? Not sure if the local integration tests catch the integration issues between Protobuf, ML-IO, and PyArrow or not.

Will do, doesn't look like local tests cover everything.

dewan-c:
> Was this branch ever properly integrated with CodeBuild? As in, is there still a follow-up change required to make the builds successful?

There are followups, as mentioned in the description, I'll add the buildspec. I setup Webhook earlier. I don't think it was properly integrated with Codebuild earlier.

amzn-choeric:
Oh okay. I was wondering if there was some other build you were looking at when you mentioned the cause of the failure since I believe the one I looked at just would not even execute properly the way it was configured.

dewan-c:
Ah ya, I was looking at the existing 1.0-1 build project. It appears there were build a couple of months ago and I was referring to the last build, which resulted in errors. I'll ping you offline to see if we're on the same page.

mabunday (approved):

amzn-choeric (approved):

PR #348, opened 2022-12-22
MMS UseContainerSupport Flag

amzn-choeric:
*Issue #, if available:*

*Description of changes:*
The Java-based application portion of MMS is unable to detect the actual number of CPU cores due to a back-ported change in Java. Disabling the container support flag that is now enabled by default restores the original behavior, which allows in proper computation of thread allocation and what-not.

I am not disabling by default as this will have impacts to current production workloads. In future versions/releases, however, this should be disabled by default.

*Testing:*
Ran integration tests against the generated image and confirmed the expected number of CPUs in a pre-prod environment.

### DIFF
--- a/src/sagemaker_xgboost_container/serving_mms.py
+++ b/src/sagemaker_xgboost_container/serving_mms.py
@@ -102,6 +102,12 @@ def _set_mms_configs(is_multi_model, handler):
     # JVM configurations for MMS, exposed to users as env vars
     _set_default_if_not_exist("SAGEMAKER_MAX_HEAP_SIZE", str(max_heap_size) + "m")
     _set_default_if_not_exist("SAGEMAKER_MAX_DIRECT_MEMORY_SIZE", os.environ["SAGEMAKER_MAX_HEAP_SIZE"])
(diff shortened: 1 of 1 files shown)


mabunday (approved):
Thanks for the change. Let's create a backlog item on our side to enable this by default with 1.6 maybe.

src/sagemaker_xgboost_container/serving_mms.py
```diff
@@ -102,6 +102,12 @@ def _set_mms_configs(is_multi_model, handler):
     # JVM configurations for MMS, exposed to users as env vars
     _set_default_if_not_exist("SAGEMAKER_MAX_HEAP_SIZE", str(max_heap_size) + "m")
     _set_default_if_not_exist("SAGEMAKER_MAX_DIRECT_MEMORY_SIZE", os.environ["SAGEMAKER_MAX_HEAP_SIZE"])
+    _set_default_if_not_exist("SAGEMAKER_MAX_DIRECT_MEMORY_SIZE", os.environ["SAGEMAKER_MAX_HEAP_SIZE"])
```
dewan-c, line 105: Duplicate line.

PR #359, opened 2022-12-29
Unused Import Fix

amzn-choeric:
*Issue #, if available:*

*Description of changes:*
Sort of odd flake issue that suddenly started failing the build. Was not even able to reproduce it myself with my recent changes until I rebuilt the entire thing. May be a dependency-related issue as no recent changes were made surrounding the area it is complaining about, and it built perfectly fine with this change in the last successful build a while ago as well.

*Testing:*
Ran flake8 locally with the test container.

### DIFF
--- a/src/sagemaker_xgboost_container/encoder.py
+++ b/src/sagemaker_xgboost_container/encoder.py
@@ -19,7 +19,7 @@ import json
 import logging
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


mabunday:
Doesn't look like we fix the flake8 version in the test dependencies, so that may be the reason.

mabunday (approved):

PR #366, opened 2023-02-02
Multi-GPU training 

dewan-c:
This is the very basic setup to run a training job in algorithm mode with Dask. There are more improvements and integration needed, which I'll create tasks for. 

Rev 2: 
* Addressed comments, refactored, added tests.
* Added some test data:
* * CSV files were added just to test that the read captures multiple files.
* * The Parquet files were added as Dask can't work with the current ones. Created a task internally to better handle these file issue.

*Testing:*
* Unit tests
* Ran a training job in script mode.


src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+
+def run_training_with_dask(
+        hyperparameters: {},
+        train_path: str,
+        validation_path: str,
+        model_dir: str,
+        content_type: str,
+        sm_hosts: [str],
+        current_host: str,
+        num_gpus: int,
```
mabunday, line 103:
```suggestion
        hyperparameters: Dict,
        train_path: str,
        validation_path: str,
        model_dir: str,
        content_type: str,
        sm_hosts: List[str],
        current_host: str,
        num_gpus: int,
```

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+                watchlist.append((dtrain, 'train'))
+
+                if validation_path is not None:
+                    X_valid, y_valid = _read_data(client, validation_path, content_type)
+                    dvalid = xgb.dask.DaskDMatrix(client, X_valid, y_valid)
+                    watchlist.append((dvalid, 'validation'))
+
+            except Exception as e:
+                raise exc.UserError("Failed to load data with exception:\n{}".format(e))
+                return
```
mabunday, line 141:
```suggestion
```

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+                                        dtrain,
+                                        num_boost_round=hyperparameters["num_round"],
+                                        evals=watchlist)
+                booster = output['booster']
+
+                logging.info("Training complete. Saving model...")
+                booster.save_model(os.path.join(model_dir, MODEL_NAME))
+            except Exception as e:
+                exception_prefix = "XGB train call failed with exception"
+                raise exc.AlgorithmError("{}:\n {}".format(exception_prefix, str(e)))
```
mabunday, line 157:
```suggestion
                raise exc.AlgorithmError(f"XGB train call failed with exception:\n {str(e)}")
```

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+
+            logging.info("Terminating cluster...")
+
+    else:
+        # Do not exit till the job is done.
+        while True:
```
mabunday, line 163: Do we need to break out of this loop?

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+            alive_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            alive_check = alive_socket.connect_ex(scheduler)
+            if alive_check == 0:
+                pass
+            else:
+                logging.info("Received a shutdown signal from scheduler. Exiting...")
+                sys.exit(0)
+            alive_socket.close()
+            time.sleep(2)
+    return
```
mabunday, line 174:
```suggestion
```

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -117,6 +117,12 @@ def get_validated_dmatrices(
 
     return train_dmatrix, val_dmatrix, train_val_dmatrix
 
+# TODO: Add data distributionType check here.
+def eligible_for_dask_run(tree_method_selected: str, num_gpus: int, is_pipe: bool, content_type: str) -> bool:
```
mabunday, line 121: Where do we call this?

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def _start_daemons(scheduler_host_ip: str, current_host_ip: str):
+    cmd_start_scheduler = os.path.join(DASK_PATH, "dask-scheduler")
+    cmd_start_worker = os.path.join(DASK_PATH, "dask-cuda-worker")
+    schedule_conn_string = "tcp://{}:{}".format(scheduler_host_ip, SCHEDULER_PORT)
+
+    # We can set this based on number of vCPUs in the future.
+    # So far haven't seen improvement with > 4, during testing.
+    num_threads_per_worker_process = str(4)
```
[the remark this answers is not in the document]
dewan-c: I wanted to keep this here since this a dask only constant. 
Also removed the silly str conversion in last commit.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+    # Dask cuda worker API doc: https://docs.rapids.ai/api/dask-cuda/nightly/api.html
+    worker_cli_command = [cmd_start_worker, schedule_conn_string, "--no-dashboard",
+                          "--nthreads", num_threads_per_worker_process]
+    # Dask distributed scheduler API doc: https://docs.dask.org/en/stable/deploying-cli.html
+    scheduler_cli_command = [cmd_start_scheduler, "--no-dashboard"]
+    if scheduler_host_ip == current_host_ip:
+        Popen(scheduler_cli_command)
+    Popen(worker_cli_command)
```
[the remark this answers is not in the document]
dewan-c: I think Popen is fine as we don't want to block and move on to start other processes. There might be variance in the startup time for each daemon, but we can account for that by waiting later on.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def _start_daemons(scheduler_host_ip: str, current_host_ip: str):
+    cmd_start_scheduler = os.path.join(DASK_PATH, "dask-scheduler")
+    cmd_start_worker = os.path.join(DASK_PATH, "dask-cuda-worker")
+    schedule_conn_string = "tcp://{}:{}".format(scheduler_host_ip, SCHEDULER_PORT)
+
+    # We can set this based on number of vCPUs in the future.
+    # So far haven't seen improvement with > 4, during testing.
+    num_threads_per_worker_process = "4"
```
amzn-choeric, line 46: Have you had a chance to check to see the difference when this is set to 1? It seems as though we should not really expect significant performance gains from values greater than 1 since they would all be locked on the same stream.  What really seems to matter is ensuring that we are creating a process per GPU as available. I believe this is implicit within the worker implementation, but do you know if this is true?

src/sagemaker_xgboost_container/algorithm_mode/train.py
dewan-c: In line 185 below. I commented out the actual call, till we're ready to enable the usage.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+def _get_host_ip(host_name: str) -> str:
+    ip_wait_time = 5
+    counter = 0
+    host_ip = ""
+    from socket import gaierror
+    while counter < ip_wait_time and host_ip == "":
+        try:
+            host_ip = socket.gethostbyname(host_name)
+            break
+        except gaierror:
```
amzn-choeric, line 66: Is this related to waiting for hosts to initialize or something else? I am wondering if this is really necessary since we do not really do this check for standard distributed jobs. We primarily just call `socket.gethostbyname` for each workers to check to see if each host is initialized.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+            counter += 1
+            time.sleep(1)
+
+    if counter == ip_wait_time and host_ip == "":
+        raise Exception(
```
amzn-choeric, line 71: Sort of related to my comment above of if this is necessary. If we do end up throwing an exception, would it be better to throw a more specific/internalized one?

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+
+    features, labels = client.persist(
+        [features, labels]
+    )
+    wait([features, labels])
+
+    return features, labels
+
+
+def run_training_with_dask(
```
amzn-choeric, line 95: Within this module, would it be better to group things together under a class or separate more common utilities out into a separate module where reusability may apply?

Within this function, would it be possible to separate out data loading and execution?

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -170,6 +172,33 @@ def sagemaker_train(
...
+        logging.info("Will run distributed GPU training with Dask.")
+        run_training_with_dask(validated_train_config,
+                               train_path,
+                               val_path,
+                               model_dir,
+                               sm_hosts,
+                               sm_current_host,
+                               num_gpus)
+        return
+    '''
```
amzn-choeric, line 195: At minimum (when uncommented), this should probably be moved out of this method.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
dewan-c: Good point, I don't have the numbers for comparison of 1 vs more. I can run an existing job to see.
And yes, we're creating a worker process per GPU, that's already handled by dask-cuda.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
dewan-c: Ah ya, I see that in the tracker code. Will remove the explicit wait and just get the id.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+    if counter == ip_wait_time and host_ip == "":
+        raise Exception(
+            "Exceeded max wait time of {}s for hostname resolution for host {}.".format(ip_wait_time, host_name)
+        )
+    return host_ip
+
+
+def _read_data(client: Client, local_path: str, content_type: str) -> (DataFrame, Series):
+    if content_type == "csv":
+        dataframe = dask_dataframe.read_csv(local_path + "/*.csv", header=None)
```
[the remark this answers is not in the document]
dewan-c: Ya, trailing slashes will cause issue, although I haven't seen that yet. Will change to use os.path.join.
In our documentation we specify that CSV shouldn't have a header record. https://docs.aws.amazon.com/sagemaker/latest/dg/xgboost.html

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
dewan-c: Actually, break would be more natural.

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
```diff
@@ -0,0 +1,119 @@
...
+            logging.info("Data load complete. Starting training...")
+
+            try:
+                output = xgb.dask.train(
+                    client, hyperparameters, dtrain, num_boost_round=hyperparameters["num_round"], evals=watchlist
+                )
+                booster = output["booster"]
+
+                logging.info("Training complete. Saving model...")
+                booster.save_model(os.path.join(model_dir, MODEL_NAME))
```
dewan-c, line 100: Leaving this here for now, but we should refactor to use a common function to save models, for all frameworks.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
```diff
@@ -0,0 +1,174 @@
...
+import os
+import socket
+from subprocess import Popen
+import sys
+import time
+
+import dask.dataframe as dask_dataframe
+from dask.dataframe import DataFrame
+from dask.dataframe import Series
+from dask.distributed import Client, wait
```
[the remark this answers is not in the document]
dewan-c: Ran isort, black and flake8.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
dewan-c: I saw about a 30 sec improvement with 4 over 1. The total runtime was 16mins. Removed this arg to use the default 1 thread.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
dewan-c: Refactored this and now throwing PlatformError.

src/sagemaker_xgboost_container/algorithm_mode/dask_training.py
dewan-c: Did some refactoring.

dewan-c:
@mabunday @amzn-choeric This should address all the major feedbacks. Thanks for the great reviews so far!

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
```diff
@@ -0,0 +1,119 @@
...
+
+                watchlist.append((dtrain, "train"))
+
+                if validation_path is not None:
+                    X_valid, y_valid = load_data_into_memory(client, validation_path, content_type)
+                    dvalid = xgb.dask.DaskDMatrix(client, X_valid, y_valid)
+                    watchlist.append((dvalid, "validation"))
+
+            except Exception as e:
+                raise exc.UserError(f"Failed to load data with exception:\n{e}")
```
mabunday, line 89: Some general comments that don't necessarily need to be addressed, but should if feasible:

1. Try... Except blocks should be as specific and concise as possible. For example here if we're trying to catch exception while loading data, it seems like the try should only encompass the `load_data_into_memory` call. Same below with the train call.
2. Wherever we use a UserError exception we should try and make the error message actionable, to the best of our ability.
3.

src/sagemaker_xgboost_container/distributed_gpu/dask_cluster_utils.py
```diff
@@ -0,0 +1,40 @@
...
+import socket
+from subprocess import Popen
+
+from sagemaker_algorithm_toolkit.exceptions import PlatformError
+
+SCHEDULER_EXEC_PATH = "/miniconda3/bin/dask-scheduler"
+CUDA_WORKER_EXEC_PATH = "/miniconda3/bin/dask-cuda-worker"
+
+
+def start_daemons_in_current_instance(scheduler_conn_string: str, is_scheduler_host: bool):
```
mabunday, line 23: I'm still a little confused about this function and why we use `Popen` instead of `call`. Below, we call this function and then wait `WAIT_TIME_DAEMONS_STARTUP_SEC`, presumably for the scheduler and worker to come online, right? I think if we used `call` here, which is blocking, this function wouldn't return until the scheduler and worker came online, which I think is exactly what we're trying to wait for with the sleep.

mabunday (changes requested):

src/sagemaker_xgboost_container/distributed_gpu/dask_data_utils.py
```diff
@@ -0,0 +1,50 @@
...
+from sagemaker_xgboost_container.data_utils import CSV, PARQUET
+
+
+def _read_data(local_path: str, content_type: str) -> (DataFrame, Series):
+    if content_type == CSV:
+        dataframe = dask_dataframe.read_csv(os.path.join(local_path, "*.csv"), header=None)
+    elif content_type == PARQUET:
+        dataframe = dask_dataframe.read_parquet(local_path)
+    else:
+        raise ValueError(f"Unexpected content type '{content_type}'. Supported content types are CSV and PARQUET.")
```
mabunday, line 29: Should this be a customer error instead?

src/sagemaker_xgboost_container/distributed_gpu/dask_cluster_utils.py
dewan-c: Good point. My reasoning for using Popen and explicit wait seems to be flawed now. Call seems simpler. Will update.

dewan-c: Actually, wouldn't call just block since the commands will be starting long running processes? Unless I'm understanding the documentation wrong it only exists once the command completes: https://docs.python.org/3/library/subprocess.html#subprocess.call

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
dewan-c: Got it. I'll move the data load functionality into the util module and raise exception from there.

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
```diff
@@ -0,0 +1,119 @@
...
+            scheduler = (scheduler_host_ip, int(SCHEDULER_PORT))
+            alive_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            alive_check = alive_socket.connect_ex(scheduler)
+            if alive_check == 0:
+                pass
+            else:
+                logging.info("Received a shutdown signal from scheduler. Exiting...")
+                break
+            alive_socket.close()
+            time.sleep(WORKER_STAY_ALIVE_CHECK_FREQ_SEC)
```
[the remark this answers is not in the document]
dewan-c: If we want to keep a socket alive and call connect_ex we'll have to handle "already connected" 106 errors.

src/sagemaker_xgboost_container/distributed_gpu/dask_cluster_utils.py
mabunday: Ah you're right - I misunderstood what those calls would return.

Popen should be correct here but we might want to think about whether there's a better way to check whether the scheduler and workers are online than a constant wait. Is there any way we can do a health check of sorts?

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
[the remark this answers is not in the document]
mabunday: Hmm, so if I'm following correctly on line 117 there's an execution path where we want to break out of the while loop, but keep `alive_socket` open? When would we close the socket then?

src/sagemaker_xgboost_container/distributed_gpu/dask_cluster_utils.py
dewan-c: I see that the [Client](https://distributed.dask.org/en/stable/api.html#distributed.Client) object has a timeout param. Not sure what the default timeout value is, but that param + wait_for_workers should be enough. Will test this out.

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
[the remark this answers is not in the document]
dewan-c: In this case it wouldn't matter whether we close the connection or not, since that'd be the end of the program. This `else` portion is purely just to keep the programs in non-scheduler hosts from exiting early.
Your comment raises a good question though, whether opening a new connection every 2s will cause any issues, especially when we get upto a large number of hosts. For now I'd prefer to move forward with a longer check period, and we can revisit this as follow-up.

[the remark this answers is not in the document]
mabunday: OK, in that case I think this should work then and aligns with the original functionality:

```python3
         while True:
            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as alive_socket:
                scheduler = (scheduler_host_ip, int(SCHEDULER_PORT))
                alive_check = alive_socket.connect_ex(scheduler)
                if alive_check != 0:
                    logging.info("Received a shutdown signal from scheduler. Exiting...")
                    break
            time.sleep(WORKER_STAY_ALIVE_CHECK_FREQ_SEC)
```

src/sagemaker_xgboost_container/distributed_gpu/dask_cluster_utils.py
dewan-c: No, that doesn't work. Without explicit wait some workers start up before the scheduler and these workers don't seem to retry connecting to the scheduler. So the cluster ends up with very few workers. 
The other option seems to be to create a Client to check if scheduler is up, before starting the workers. However, I'd like to backlog this. Lmk if there are concerns.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -170,6 +165,15 @@ def sagemaker_train(
         elif not is_pipe:
             # Check if there is potential data redundancy between training and validation sets
             check_data_redundancy(train_path, val_path)
+
```
dewan-c, line 168: Removed the unused call to Dask training for now.

trivialfis:
Hi @dewan-c @mabunday , I help maintain the XGBoost library. We added a new type `DaskQuantileDMatrix` designed for the hist tree method. It uses significantly less memory when compared to the `DaskDMatrix`. Is there a way that we can help?

PR #370, opened 2023-02-12
Fix dataframe operation after read

dewan-c:
`dataframe.columns.difference`  returns a sorted list of column names, unless explicitly told not to sort. This messes up the feature order, if the dataframe column names were not sorted already.
Using drop() instead of modifying usage of `columns.difference`, as it looks cleaner.

### DIFF
--- a/src/sagemaker_xgboost_container/distributed_gpu/dask_data_utils.py
+++ b/src/sagemaker_xgboost_container/distributed_gpu/dask_data_utils.py
@@ -32,7 +32,7 @@ def _read_data(local_path: str, content_type: str) -> (DataFrame, Series):
 
     target_column = dataframe.columns[0]
     labels = dataframe[target_column]
-    features = dataframe[dataframe.columns.difference([target_column])]
+    features = dataframe.drop(target_column, axis=1)
 
     return features, labels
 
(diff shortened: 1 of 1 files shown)


src/sagemaker_xgboost_container/distributed_gpu/dask_data_utils.py
```diff
@@ -32,7 +32,7 @@ def _read_data(local_path: str, content_type: str) -> (DataFrame, Series):
 
     target_column = dataframe.columns[0]
     labels = dataframe[target_column]
-    features = dataframe[dataframe.columns.difference([target_column])]
+    features = dataframe.drop(target_column, axis=1)
```
amzn-choeric, line 35: Do you know if this will copy if the features and be a potential memory-related concern considering how we expect these to typically be massive datasets?

dewan-c: Hmm ya, looks like drop creates a copy.

dewan-c: Thanks, now changed to just pick all but the first column name.

amzn-choeric (approved):

PR #371, opened 2023-02-13
remove buildspec from repo

haixiw:
*Description of changes:*
removing buildspecs from github
bump up versions of some dependencies to fix the codebuild.
Updated dependencies are consistent with versions in 1.3-1.

Ref: https://github.com/aws/sagemaker-xgboost-container/pull/192


malav-shastri (approved):

amzn-choeric:
This PR does not appear to be just a removal. Are we sure all these dependency bumps are safe?

haixiw:
> This PR does not appear to be just a removal. Are we sure all these dependency bumps are safe?

we can find these dependencies has been updated to same versions in 1.3-1 in the PR below,  so I would say it wouldn't cause some regression instantly. I will run full integ tests against test image before merging the PR.
Ref: https://github.com/aws/sagemaker-xgboost-container/pull/192

PR #374, opened 2023-02-14
Dask Flow Integration

dewan-c:
This builds on top of the dask branch. The changes here are adding Dask training call to algorithm mode:
* Adding early stopping and custom metric support to Dask training.
* Adding a HP `use_dask_gpu_training ` for customers to trigger training with Dask.
* Validations to check training config is suitable for Dask. Show relevant messages otherwise.
* As commented in the code, some code blocks are duplicates and will be reworked later, due to time constraints atm.
* Removed explicit data loading in the cluster, as I saw about 15% more peak memory usage, compared to when letting the underlying DaskDMatrix code load the data. The only downside I can think of is possibly confusing error message if there are any data load failures.
* Note that custom metric(feval) arg increases the training time by considerable amount for training with large data. It's used when `eval_metric` is specified by the user or tuning metric is specified for HPO. Looking at the underlying code it deals with DMatrix directly, which could be the reason for the slowdown. It doesn't look like we're doing anything wrong.

*Testing:*
* Ran unit tests.
* Ran integ tests to confirm current code flow is uneffected.
* Ran notebook training test with Dask flag, and also with early stopping rounds.

### DIFF
--- a/src/sagemaker_xgboost_container/algorithm_mode/train.py
+++ b/src/sagemaker_xgboost_container/algorithm_mode/train.py
@@ -12,7 +12,6 @@
 # language governing permissions and limitations under the License.
 import logging
 import os
-import signal
 
 import numpy as np
 import xgboost as xgb
@@ -20,17 +19,19 @@ from sklearn.model_selection import RepeatedKFold, RepeatedStratifiedKFold
 
 from sagemaker_algorithm_toolkit import exceptions as exc
 from sagemaker_algorithm_toolkit.channel_validation import Channel
-from sagemaker_xgboost_container import checkpointing, distributed
+from sagemaker_xgboost_container import distributed
 from sagemaker_xgboost_container.algorithm_mode import channel_validation as cv
 from sagemaker_xgboost_container.algorithm_mode import hyperparameter_validation as hpv
 from sagemaker_xgboost_container.algorithm_mode import metrics as metrics_mod
 from sagemaker_xgboost_container.algorithm_mode import train_utils
-from sagemaker_xgboost_container.callback import add_debugging
-from sagemaker_xgboost_container.constants.sm_env_constants import SM_OUTPUT_DATA_DIR
+from sagemaker_xgboost_container.callback import add_debugging, get_callbacks
+from sagemaker_xgboost_container.constants.sm_env_constants import (
+    SM_NUM_GPUS,
+    SM_OUTPUT_DATA_DIR,
+)
 from sagemaker_xgboost_container.constants.xgb_constants import (
     CUSTOMER_ERRORS,
     MODEL_NAME,
-    XGB_MAXIMIZE_METRICS,
 )
 from sagemaker_xgboost_container.data_utils import (
     check_data_redundancy,
@@ -39,30 +40,12 @@ from sagemaker_xgboost_container.data_utils import (
     get_size,
     validate_data_file_path,
 )
+from sagemaker_xgboost_container.distributed_gpu import distributed_gpu_training
 from sagemaker_xgboost_container.prediction_utils import ValidationPredictionRecorder
 
 logger = logging.getLogger(__name__)
 
-
-def add_sigterm_handler(model_dir, is_master):
-    """Stop training and cleanup model directory when SIGTERM is received.
-
-    Model directory is only cleaned if is_master is True. Otherwise program terminates.
-
-    :param model_dir: Directory where model is saved
-    :param is_master: True if single node training, or the current node is the master node in distributed training
-    """
-
-    def _terminate():
-        os._exit(0)
-
-    def _cleanup_files(signo, frame):
-        if is_master:
-            train_utils.cleanup_dir(model_dir, MODEL_NAME)
-
-        _terminate()
-
-    signal.signal(signal.SIGTERM, _cleanup_files)
+DOCUMENTATION_LINK = "https://docs.aws.amazon.com/sagemaker/latest/dg/xgboost.html"
 
 
 def get_validated_dmatrices(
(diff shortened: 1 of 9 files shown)


src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
```diff
@@ -21,22 +21,53 @@
...
 WORKER_STAY_ALIVE_CHECK_FREQ_SEC = 10
+SUPPORTED_TRAINING_CONTENT_TYPES = {CSV, PARQUET}
+
+NON_GPU_ERROR_MSG = "Dask training is only available for `gpu_hist` training on GPU instances."
+PIPE_MODE_ERROR_MSG = "Dask training is not supported for pipe mode input."
+INPUT_FORMAT_ERROR_MSG = "Dask training is only supported for CSV and Parquet input."
+NOT_REPLICATED_ERROR_MSG = "Dask distributed training requires FullyReplicated data."
+
+
+def check_if_all_conditions_met(
```
mabunday, line 57:
```suggestion
def validate_multi_gpu_configuration(
```

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -169,50 +152,82 @@ def sagemaker_train(
...
+                args=train_args,
+                include_in_training=(train_dmatrix is not None),
+                hosts=sm_hosts,
+                current_host=sm_current_host,
+                update_rabit_args=True,
+            )
+        elif num_hosts == 1:
+            if train_dmatrix:
+                if validation_channel:
+                    if not val_dmatrix:
```
mabunday, line 222:
```suggestion
                if validation_channel and if not val_dmatrix:
```

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
```diff
@@ -21,22 +21,53 @@
...
 
 SCHEDULER_PORT = "8786"
 WAIT_FOR_ALL_WORKERS_TIMEOUT_SEC = 20
 WORKER_STAY_ALIVE_CHECK_FREQ_SEC = 10
+SUPPORTED_TRAINING_CONTENT_TYPES = {CSV, PARQUET}
+
+NON_GPU_ERROR_MSG = "Dask training is only available for `gpu_hist` training on GPU instances."
+PIPE_MODE_ERROR_MSG = "Dask training is not supported for pipe mode input."
+INPUT_FORMAT_ERROR_MSG = "Dask training is only supported for CSV and Parquet input."
+NOT_REPLICATED_ERROR_MSG = "Dask distributed training requires FullyReplicated data."
```
mabunday, line 54: Just a thought, but would it be a better user experience to check all the conditions in one-shot and let them know which ones they don't meet? I'm imagining a scenario where the user doesn't read the docs closely and incrementally tries to address each issue one by one, resulting in a failure each time.

src/sagemaker_xgboost_container/algorithm_mode/train.py
```diff
@@ -169,50 +152,82 @@ def sagemaker_train(
...
+            checkpoint_dir=checkpoint_dir,
+        )
+        if num_hosts > 1:
+            # Wait for hosts to find each other
+            logging.info("Distributed node training with {} hosts: {}".format(num_hosts, sm_hosts))
+            distributed.wait_hostname_resolution(sm_hosts)
+            if not train_dmatrix:
+                logging.warning(
+                    "Host {} does not have data. Will broadcast to cluster and will not be used in distributed"
+                    " training.".format(sm_current_host)
```
mabunday, line 209: For these files, feel free to install and run `pyupgrade --py38-plus`

mabunday (changes requested):

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
dewan-c: Updated to `validate_gpu_train_configuration`.

src/sagemaker_xgboost_container/algorithm_mode/train.py
dewan-c: Updated.

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
dewan-c: Good point. Modified.

src/sagemaker_xgboost_container/algorithm_mode/train.py
dewan-c: I'll run for the next PR to keep this one from having more changes.

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
```diff
@@ -21,23 +21,56 @@
...
-
+SUPPORTED_TRAINING_CONTENT_TYPES = {CSV, PARQUET}
+
+NON_GPU_ERROR_MSG = "Dask training is only available for `gpu_hist` training on GPU instances."
+PIPE_MODE_ERROR_MSG = "Dask training is not supported for pipe mode input. Please use File mode."
+INPUT_FORMAT_ERROR_MSG = "Dask training is only supported for CSV and Parquet input."
+NOT_REPLICATED_ERROR_MSG = "Dask distributed training requires FullyReplicated data."
+
+
+def validate_gpu_train_configuration(
```
mabunday, line 56: I think it'd be better to still have this function raise a UserError because then we can catch that in the unit tests as well, but feel free to leave as is.

src/sagemaker_xgboost_container/distributed_gpu/distributed_gpu_training.py
```diff
@@ -78,16 +111,63 @@ def run_training_with_dask(
...
+
             try:
                 output = xgb.dask.train(
-                    client, hyperparameters, dtrain, num_boost_round=hyperparameters["num_round"], evals=watchlist
+                    client=client,
+                    params=hyperparameters,
+                    dtrain=dtrain,
+                    num_boost_round=num_round,
+                    evals=watchlist,
+                    feval=configured_feval,
```
dewan-c, line 169: Note that I've found out that this custom metric arg increases the training time by considerable amount for training with large data. It's used when `eval_metric` is specified by the user or tuning metric is specified for HPO. Looking at the underlying code it deals with DMatrix directly, which could be the reason for the slowdown. It doesn't look like we're doing anything wrong.

mabunday (approved):

PR #377, opened 2023-02-17
Bump smdebug version to 1.0.29

malav-shastri:
*Issue #, if available:*
*Description of changes:*
- Debugger team released a fix for supporting XgBoost report and other debugger functionality for XgBoost version 1.3 and beyond - https://github.com/awslabs/sagemaker-debugger/pull/648
- Bumping the version of smdebug to 1.0.29 to patch the fix in XgBoost container 1.3-1
- Made related changes to support the new smdebug version
- Related PR -  https://github.com/awslabs/sagemaker-debugger/pull/650

*Testing:*
- Unit test
- Flake8
- Integration tests
- Tested notebook with the updated image https://github.com/aws/amazon-sagemaker-examples/blob/main/sagemaker-debugger/xgboost_training_report/higgs_boson_detection.ipynb

### DIFF
--- a/docker/1.3-1/final/Dockerfile.cpu
+++ b/docker/1.3-1/final/Dockerfile.cpu
@@ -9,8 +9,10 @@ ARG SAGEMAKER_XGBOOST_VERSION
 # Install dependencies #
 ########################
 COPY requirements.txt /requirements.txt
-RUN python3 -m pip install -r /requirements.txt && rm /requirements.txt && \
-    python3 -m pip install git+https://github.com/awslabs/sagemaker-debugger.git@1.0.28
+RUN python3 -m pip install -r /requirements.txt && rm /requirements.txt
+
+# Install smdebug from source
+RUN python3 -m pip install git+https://github.com/awslabs/sagemaker-debugger.git@1.0.28
 
 
 ###########################
--- a/docker/1.3-1/final/Dockerfile.cpu
+++ b/docker/1.3-1/final/Dockerfile.cpu
@@ -12,7 +12,7 @@ COPY requirements.txt /requirements.txt
 RUN python3 -m pip install -r /requirements.txt && rm /requirements.txt
 
 # Install smdebug from source
-RUN python3 -m pip install git+https://github.com/awslabs/sagemaker-debugger.git@1.0.28
+RUN python3 -m pip install git+https://github.com/awslabs/sagemaker-debugger.git@1.0.29
 
 
 ###########################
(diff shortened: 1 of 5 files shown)


docker/1.3-1/final/Dockerfile.cpu
```diff
@@ -11,6 +11,10 @@ ARG SAGEMAKER_XGBOOST_VERSION
 COPY requirements.txt /requirements.txt
 RUN python3 -m pip install -r /requirements.txt && rm /requirements.txt
 
+# Install smdebug from source
+RUN python3 -m pip install git+https://github.com/awslabs/sagemaker-debugger.git@1.0.29
```
dewan-c, line 15: Curious, why are we installing from branch here if it's available through pip?

src/sagemaker_xgboost_container/callback.py
```diff
@@ -37,9 +17,13 @@ def add_debugging(callbacks, hyperparameters, train_dmatrix,
...
+        hook = Hook.hook_from_config(json_config_path)
+        hook.hyperparameters = hyperparameters
+        hook.train_data = train_dmatrix
+        if val_dmatrix is not None:
+            hook.validation_data = val_dmatrix
+        callbacks.append(hook)
         logging.info("Debug hook created from config")
     except Exception as e:
         logging.debug("Failed to create debug hook", e)
-    else:
```
dewan-c, line 44: Nit - this else is now redundant.

docker/1.3-1/final/Dockerfile.cpu
malav-shastri: so this new version is not available through pip, according to them they haven't released wheels recently and their pipeline for that is blocked. They are not sure when they would be able to resolve that and suggested me to directly install it from source as almost all of their dependencies are installing from source only. To unblock this issue for time being if there are no major cons to this approach maybe we can treat it as a temporary work around and can request them to release a new wheel eventually.

src/sagemaker_xgboost_container/callback.py
malav-shastri: umm so currently I am not using it right?

docker/1.3-1/final/Dockerfile.cpu
dewan-c: Ok, makes sense, thanks for the explanation.

src/sagemaker_xgboost_container/callback.py
dewan-c: Why keep unused code? I meant removing this else altogether would work just the same.

dewan-c (approved):

src/sagemaker_xgboost_container/callback.py
dewan-c: Ak it's my bad. I didn't notice that return was moved under try. Fyi- you don't need the `return` either.

PR #396, opened 2023-06-15
Fix: load training data from all sub dirs

haixiw:
*Issue #, if available:*
https://answers.amazon.com/posts/362803

Previously if customer set the input files channel to be `s3://bucket/sagemaker/train/` (file mode)
it will only load the files directly under this path or a random sub dir based on workflow here

https://github.com/aws/sagemaker-xgboost-container/blob/master/src/sagemaker_xgboost_container/data_utils.py#L557
https://github.com/aws/sagemaker-xgboost-container/blob/master/src/sagemaker_xgboost_container/data_utils.py#L638-L647

With this change, the behavior of loading files will be consistent with pipe mode, which is that all the sub folders under the root dir customer specified will be traversed and all the files found there will be loaded.

*Description of changes:*
1. Add a helper function to recursively creating symlinks from a folder to another folder
2. Using the above helper function to create symlinks from all sub folders under root dir user specified to a temp folder. (XGB could load all files under the same dir)
3. In order to prevent name collision as all the files are linked to the same dir, I added a suffix for the new filename in temp dir with the hash of the source file's path.

*Testing:*
unit tests,
Tested with Notebook : https://haixiw-test.notebook.us-west-2.sagemaker.aws/notebooks/XGBoost_abalone/xgboost_abalone.ipynb


mabunday:
Hmm my only concern here is that it changes behavior for existing customers. Can you also take a look at the PR that added the function in question and see if there's anything we need to account for here? https://github.com/aws/sagemaker-xgboost-container/pull/158/files

malav-shastri:
Thanks for this PR, apart from above just curious to know, as discussed in our sync were we actually able to find similar logic in 1P closed source containers(AIAlgorithmsSDK) where this is already being supported? And whether this fix is based on that to maintain consistency(although that's not needed but just to get the understanding about it)

haixiw:
> Hmm my only concern here is that it changes behavior for existing customers. Can you also take a look at the PR that added the function in question and see if there's anything we need to account for here? https://github.com/aws/sagemaker-xgboost-container/pull/158/files

Thanks for the headsup. I think theoretically this change won't affect the previous workflow with cross validation after revisited  https://github.com/aws/sagemaker-xgboost-container/pull/158/files.  There's a local integ test for cross validation here https://github.com/aws/sagemaker-xgboost-container/blob/master/test/integration/local/test_kfold.py, also I will sync with AP oncall to see if they could perform their integ tests on some test images.

haixiw:
> Thanks for this PR, apart from above just curious to know, as discussed in our sync were we actually able to find similar logic in 1P closed source containers(AIAlgorithmsSDK) where this is already being supported? And whether this fix is based on that to maintain consistency(although that's not needed but just to get the understanding about it)

Thanks for the comment! I haven't looked into that closely but let's keep this discussion through slack as we don't want expose much SDK stuff here.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -528,33 +528,50 @@ def _get_pipe_mode_files_path(data_path: Union[List[str], str]) -> List[str]:
     return files_path
 
 
+def _make_symlinks_for_files_under_a_folder(dest_path: str, data_path: str):
+    if (not os.path.exists(dest_path)) or (not os.path.exists(data_path)):
+        raise exc.AlgorithmError("Unable to create symlinks as {data_path} or {dest_path} doesn't exist ")
```
mabunday, line 533: Is this meant to be an f-string?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -528,33 +528,50 @@ def _get_pipe_mode_files_path(data_path: Union[List[str], str]) -> List[str]:
...
+
+    logging.info("Making smlinks from folder {} to folder {}".format(data_path, dest_path))
```
mabunday, line 535: nit: Can use f-strings here instead of format for better readability.

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -528,33 +528,50 @@ def _get_pipe_mode_files_path(data_path: Union[List[str], str]) -> List[str]:
...
+    if os.path.isfile(data_path):
+        _make_symlink(data_path, dest_path, os.path.basename(data_path))
+        return
+
+    else:
+        for item in os.scandir(data_path):
+            if item.is_file():
+                _make_symlink(item.path, dest_path, item.name)
+            elif item.is_dir():
+                _make_symlinks_for_files_under_a_folder(dest_path, item.path)
```
mabunday, line 546: Do we need to worry about recursion depth here?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -528,33 +528,50 @@ def _get_pipe_mode_files_path(data_path: Union[List[str], str]) -> List[str]:
...
     else:
         if not os.path.exists(data_path):
             logging.info("File path {} does not exist!".format(data_path))
             return None
-        files_path = get_files_path_from_string(data_path)
+        elif os.path.isdir(data_path):
+            # traverse all sub-dirs to load all training data
+            _make_symlinks_for_files_under_a_folder(files_path, data_path)
+        elif os.path.isfile(data_path):
+            files_path = data_path
```
mabunday, line 572: Can we just call `_make_symlinks_for_files_under_a_folder ` here? Isn't it able to handle singular files:?

src/sagemaker_xgboost_container/data_utils.py
```diff
@@ -635,22 +652,11 @@ def get_size(data_path, is_pipe=False):
...
-    return files_path
-
-
-def _make_symlink(path, source_path, name, index):
-    base_name = os.path.join(source_path, f"{name}_{str(index)}")
-    logging.info(f"creating symlink between Path {source_path} and destination {base_name}")
-    os.symlink(path, base_name)
+def _make_symlink(path, source_path, name):
+    base_name = os.path.join(source_path, name)
+    file_name = base_name + str(hash(path))
```
mabunday, line 657: What happens in case of hash collisions?

src/sagemaker_xgboost_container/data_utils.py
haixiw: Nice catch, yeah, I will update

src/sagemaker_xgboost_container/data_utils.py
haixiw: sure, Thanks

src/sagemaker_xgboost_container/data_utils.py
haixiw: emmm, I was thinking about this too. I will add a depth limit of 5 here.

src/sagemaker_xgboost_container/data_utils.py
haixiw: yes, it 's able to do that..Let me refactor the info message in next commit to prevent confusion.

src/sagemaker_xgboost_container/data_utils.py
haixiw: An error will be thrown out saying there is an existing symlink between A and B. The reason for `hash(path)` is that no matter how complicated the file structure is, the path for a single file is normally unique and it won't be too long of a string when we limit the depth for recursion.

mabunday (approved):

malav-shastri:
Do we need to fix 1.3 and 1.5 as well?

malav-shastri (approved):

haixiw:
> Do we need to fix 1.3 and 1.5 as well?

Yes. This seems a general issue for all versions.

PR #452, opened 2025-08-12
update libsqlite version

bnpaws:
*Issue #, if available:*

*Description of changes:*

*Testing:*


bnpaws:
```
#18 [stage-1 10/12] RUN sqlite3 --version
#18 0.455 3.50.2 2025-06-28 14:00:48 2af157d77fb1304a74176eaee7fbc7c7e932d946bf25325e9c26c91db19e3079 (64-bit)
#18 DONE 0.6s
```

bnpaws:
for `CVE-2025-6965` in SQLite, a custom installation via wget is necessary because the ubuntu (even 24.04 has latest 37.X version); the recommended fix is in 3.50.2 which is available via wget (checked other repos). While apt-get will continue to show the older version of libsqlite3-0 (occurs because APT's package database operates independently from the actual filesystem contents) that's in the repository, our manual compilation and installation from source places the newer SQLite 3.50.2 binaries and libraries in the system directories (/usr). The system's dynamic linker (configured through ldconfig) will prioritize these newer libraries, ensuring that all applications using SQLite link against the secure 3.50.2 version. This approach allows us to patch the security vulnerability.

malav-shastri (approved):

PR #467, opened 2026-02-27
add MetricHistory and MetricsTracker with XGBoost callback support

alekseevpavel04:
*Issue #, if available:*
N/A

*Description of changes:*
The Metrics class had no way to track values over training time or detect stagnation. Added MetricHistory (per metric value log with direction-aware best tracking and rounds-without-improvement counting) and MetricsTracker (collection-level wrapper with early stopping detection). MetricsTracker.xgboost_callback() returns a TrainingCallback that pipes xgb.train() evals_log directly into the tracker each round and signals XGBoost to halt when all tunable metrics have stalled for early_stopping_rounds iterations. Both classes follow the existing style and integrate naturally with the Metric/Metrics objects already in this file.

*Testing:*
Verify MetricHistory.best_value respects MAXIMIZE vs MINIMIZE, MetricsTracker.should_stop_early only triggers after the configured stall count, and xgboost_callback feeds evals_log values correctly into each MetricHistory.

### DIFF
--- a/src/sagemaker_algorithm_toolkit/metrics.py
+++ b/src/sagemaker_algorithm_toolkit/metrics.py
@@ -58,3 +58,196 @@ class Metrics(object):
 
     def format_definitions(self):
         return [metric.format_definition() for name, metric in self.metrics.items()]
+
+
+class MetricHistory(object):
+    """Record and analyse the value of a single Metric across training iterations."""
+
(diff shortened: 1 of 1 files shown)


alekseevpavel04:
Closing this since I'm not able to follow up on it any more. Thanks for taking a look!

PR #483, opened 2026-05-11
fix: bypass dask.dataframe expression optimizer bug in DaskDMatrix cr…

Jyothirmaikottu:
…eation

dask >= 2026.1 has a bug in its expression optimizer where column comparison returns an array instead of a scalar, causing ValueError when DaskDMatrix tries to persist/optimize DataFrame expressions.

Fix: read data with pandas directly and convert to dask.array before passing to DaskDMatrix. This bypasses the dask.dataframe expression layer entirely while preserving distributed GPU training functionality.

*Issue #, if available:*

*Description of changes:*

*Testing:*

### DIFF
--- a/src/sagemaker_xgboost_container/distributed_gpu/dask_data_utils.py
+++ b/src/sagemaker_xgboost_container/distributed_gpu/dask_data_utils.py
@@ -11,11 +11,12 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
+import glob
 import os
 
-
-import dask.dataframe as dask_dataframe
-from dask.dataframe import DataFrame, Series
+import dask.array as da
+import numpy as np
+import pandas as pd
 from dask.distributed import Client
 from xgboost import dask as dxgb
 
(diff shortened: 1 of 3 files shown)


src/sagemaker_xgboost_container/distributed_gpu/dask_data_utils.py
```diff
@@ -11,50 +11,65 @@
...
+    expression optimizer bug in dask >= 2026.1.
+    """
     try:
-        dmatrix = dxgb.DaskDMatrix(client, features, labels)
+        features_np = features.values if hasattr(features, 'values') else np.asarray(features)
+        labels_np = labels.values if hasattr(labels, 'values') else np.asarray(labels)
+
+        features_da = da.from_array(features_np, chunks=features_np.shape)
+        labels_da = da.from_array(labels_np, chunks=labels_np.shape)
+        dmatrix = dxgb.DaskDMatrix(client, features_da, labels_da)
```
jinyan-li1, line 78: I'm a bit worried about a regression here - the change  reads all CSVs into the driver and wraps them as a single chunk in `da.from_array(..., chunks=features_np.shape)`. So even though it's a Dask  object, only one worker actually trains on it. On a real multi-GB customer dataset that'd either OOM the driver or be a lot slower than before.

Jyothirmaikottu: Good point, changed it to chunks. Either way, sagemaker downloads training dataset to local instance disk so if in the case of large datasets, this can be handled with [fullyreplicated](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_S3DataSource.html) distribution on multiple instances.

test/resources/versions/train.py
```diff
@@ -24,7 +24,7 @@
 sagemaker-inference==1.5.5
 scipy==1.15.0
 scikit-learn==1.8.0
-urllib3==2.4.0
+urllib3==2.7.0
```
Jyothirmaikottu, line 27: fixed cve

jinyan-li1 (approved):

